PR #150, opened 2026-04-14
Fix #127 P1: Rate-limit /device/authorize OAuth init endpoint

### LINKED ISSUE #127: Local Ktor server is bound to 0.0.0.0 and some endpoints are LAN-reachable without auth
## Problem
\`McpSession.start(port)\` calls \`embeddedServer(CIO, port = port) { ... }\` at \`core/mcp/src/commonMain/kotlin/com/rousecontext/mcp/core/McpSession.kt:65\` without passing a \`host\` argument. Ktor's \`embeddedServer\` defaults \`host\` to \`\"0.0.0.0\"\` — the server listens on every network interface, not just loopback.

On a typical phone on WiFi, this means the device is serving the MCP Ktor instance on its LAN IP:
- Same-LAN attackers can port-scan and fingerprint the device as running Rouse Context
- \`/.well-known/oauth-*\` discovery endpoints are publicly reachable
- \`/authorize\` and device-code flow initiation are public (per OAuth spec — but need not be reachable by arbitrary LAN hosts)
- \`/mcp\` is protected by Bearer (\`McpRouting.kt:513-518\`) so they can't call tools, but they CAN spam authorization requests and trigger user-facing approval UI

Same-device app-to-app reachability on loopback is a separate concern even after fixing the bind — Android apps CAN connect to each other's 127.0.0.1 sockets without special permissions.

### ~~P0 — Bind to loopback only~~ (landed in 50732a0)
\`\`\`kotlin
embeddedServer(CIO, port = port, host = "127.0.0.1") { ... }
\`\`\`
One line. Closes the entire LAN attack surface. Must land first.

### P1 — Rate-limit OAuth initiation
Audit the existing \`rateLimiter\` in \`configureMcpRouting\`: does it cover \`/authorize\` and device-code start? If not, add per-IP (on loopback that's always 127.0.0.1, so effectively per-device) rate limits on those endpoints to prevent approval-UI spam from a hostile same-device app.

### P2 — Internal shared-secret header
Since the only legitimate HTTP caller is \`McpSessionBridge\` in the same process, have the bridge inject an \`X-Internal-Token: <random-per-session-secret>\` header. Ktor route intercepts before OAuth and rejects requests missing/incorrect. Defense-in-depth against other apps on device hitting loopback.

Generate the secret in \`McpSession.start()\` (at server-start time, kept in memory, never persisted) and pass to the bridge via the already-existing injection path.

### P3 — Unix domain socket (alternative to P2)
Replace loopback TCP between bridge and Ktor with \`AF_UNIX\` (Android \`LocalSocket\`/\`LocalServerSocket\`). The socket lives in the app's private directory (UID-scoped) so other apps can't connect. Blocks same-device apps structurally.

Cost: Ktor CIO doesn't support AF_UNIX natively. Would need either a small bridge socket-to-socket shim, or replacing Ktor's listener with a custom \`ApplicationEngine\`. Significant work; only worth it if P2 proves insufficient.

## Tests
- **P0**: integration test — start McpSession, attempt connect to server's external IP, assert connection refused. Existing loopback tests still work.
- **P1**: unit test — hit \`/authorize\` repeatedly beyond rate limit, assert 429.
- **P2**: unit test — request without \`X-Internal-Token\` header → 403; with correct token → normal behavior.

## Out of scope
- This is NOT a remote-exploit CVE — the relay's mTLS and the missing domain routing outside LAN prevent anything outside WiFi range. It's a local hardening issue, visible primarily when the device is on hostile networks (coffee shop, airport, multi-tenant workplace WiFi).
- Does NOT block any current use case; fix is additive.

Monkopedia:
## Summary
- Adds rate-limiting and security-alert gating to `POST /device/authorize`, the one OAuth-init endpoint that was unprotected after #127 P0 (loopback bind).
- `/register` and `/authorize` already had identical gates; this change brings `/device/authorize` in line.
- Uses existing `RateLimiter` wired via `configureMcpRouting(rateLimiter=...)` -- no new config.

## Endpoints audit (before/after)
| Endpoint | Before | After |
| --- | --- | --- |
| POST /register | rate-limited, security-alert-gated | unchanged |
| GET /authorize | rate-limited, security-alert-gated | unchanged |
| POST /device/authorize | **unprotected** | **rate-limited, security-alert-gated** |
| POST /token | rate-limited, security-alert-gated | unchanged |
| GET /authorize/status | not gated (intentional -- polling) | unchanged |
| POST /mcp | Bearer + `mcpRateLimiter` | unchanged |

## Out of scope (tracked in #127)
- P2: internal shared-secret header between McpSessionBridge and Ktor
- P3: AF_UNIX socket replacement for loopback TCP

## Test plan
- [x] New test `device authorize endpoint returns 429 when rate limited` passes; fails before the fix
- [x] Existing `/register`, `/authorize`, `/token` rate-limit tests still pass
- [x] `:core:mcp:jvmTest :core:tunnel:jvmTest :app:testDebugUnitTest ktlintCheck` all green
- [x] detekt: 74 weighted issues -- matches pre-existing baseline

Do not merge; P2/P3 remain open in #127.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/core/mcp/src/commonMain/kotlin/com/rousecontext/mcp/core/McpRouting.kt
+++ b/core/mcp/src/commonMain/kotlin/com/rousecontext/mcp/core/McpRouting.kt
@@ -304,9 +304,14 @@ internal class McpRoutes(
 
     // Device code authorize -- no auth required
     suspend fun RoutingCall.handleDeviceAuthorize() {
+        if (rejectIfSecurityAlert()) return
         val ri = resolveIntegration()
         if (registry.providerForPath(ri) == null) {
             respond(HttpStatusCode.NotFound)
+            return
+        }
+        if (rateLimiter != null && !rateLimiter.tryAcquire("$ri/device_authorize")) {
+            respond(HttpStatusCode.TooManyRequests)
             return
         }
 
--- a/core/mcp/src/jvmTest/kotlin/com/rousecontext/mcp/core/RateLimiterTest.kt
+++ b/core/mcp/src/jvmTest/kotlin/com/rousecontext/mcp/core/RateLimiterTest.kt
@@ -140,6 +140,42 @@ class RateLimiterTest {
     }
 
     @Test
+    fun `device authorize endpoint returns 429 when rate limited`() = testApplication {
+        val registry = InMemoryProviderRegistry()
+        registry.register("health", stubProvider())
+        registry.setEnabled("health", true)
+        val tokenStore = InMemoryTokenStore()
+        val deviceCodeManager = DeviceCodeManager(tokenStore = tokenStore)
+        val clock = FakeClock()
+        val rateLimiter = RateLimiter(maxRequests = 2, windowMs = 60_000L, clock = clock)
+
+        application {
+            configureMcpRouting(
+                registry = registry,
+                tokenStore = tokenStore,
+                deviceCodeManager = deviceCodeManager,
+                hostname = "test.rousecontext.com",
+                integration = "health",
+                clock = clock,
+                rateLimiter = rateLimiter
+            )
+        }
+
+        // First two requests should succeed (not 429)
+        repeat(2) {
+            val response = client.post("/device/authorize")
+            assertTrue(
+                "Request ${it + 1} should not be 429, was ${response.status}",
+                response.status != HttpStatusCode.TooManyRequests
+            )
+        }
+
+        // Third request should be rate limited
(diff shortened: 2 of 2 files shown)


Monkopedia:
Merged via cherry-pick as 1142686d. Closing PR; #127 remains open for P2/P3.

PR #153, opened 2026-04-15
Drop Android plugin from :core modules (#129)

### LINKED ISSUE #129: Drop Android library plugin from :core modules that don't use Android APIs
### `:core:bridge` — KMP-single-target, needs source rename
Plain KMP with `jvm()` only. No Android plugin. But source lives in `src/commonMain/kotlin/` despite the module being JVM-only. Same layout inconsistency as the other modules below.

### `:core:mcp` — Android plugin + target, zero Android content
- `build.gradle.kts` applies `alias(libs.plugins.android.library)` and configures `androidTarget()`
- `src/androidMain/` contained only an empty AndroidManifest.xml (deleted in 6776aa2)
- No `android.*` imports anywhere in the module
- AGP runs AAPT, manifest merging, resource processing — all for nothing
- `src/commonMain/` contains JVM-specific code (`java.util.concurrent.*`, `synchronized(this)`) which wouldn't compile on native/JS anyway

Action: drop the Android library plugin and `androidTarget()`. Keep `jvm()` and the KMP plugin.

### `:core:tunnel` — Android plugin because of manifest permissions only
- Contains `INTERNET`, `WAKE_LOCK`, `FOREGROUND_SERVICE` permission declarations in `androidMain/AndroidManifest.xml`
- No `android.*` imports in source code
- The module is Android-classified **only** to contribute manifest permissions to consumers via manifest merging

Migrate those three permissions into `:app/src/main/AndroidManifest.xml` (or `:work`, which already declares foreground-service types and is the actual owner of `startForeground`). Then drop the Android plugin from `:core:tunnel`, matching `:core:mcp`.

## Source layout — rename `commonMain` → `jvmMain`

In every affected module (`:core:mcp`, `:core:tunnel`, `:core:bridge`), rename the source directories to reflect what the code actually is:

- `src/commonMain/kotlin/` → `src/jvmMain/kotlin/`
- `src/commonTest/kotlin/` → `src/jvmTest/kotlin/` (merge with existing `jvmTest` where present)

Rationale: these modules are KMP-structured but single-target. Code in `commonMain` currently uses `java.util.concurrent.ConcurrentHashMap`, `synchronized(this)`, `java.nio.*` — JVM-only APIs. The `commonMain` label is a lie; moving to `jvmMain` makes the JVM-specificity honest. Also updates the `sourceSets` blocks in build.gradle.kts accordingly (the KMP plugin already understands `jvmMain` natively — it's the default source set name for the `jvm()` target).

Do NOT drop the KMP plugin entirely. We keep `kotlin-multiplatform` so a future target is a plain addition, not a restructure. The only thing changing is honesty about *today*.

## Benefits
- Faster builds on every full compile (no AAPT, no manifest merging, no AGP overhead for Android-classified modules that don't need it)
- Smaller dependency cones
- Module classification and source layout match actual content
- Makes code like `java.util.concurrent.ConcurrentHashMap` visibly belong to the JVM target, not pretend commonMain code

### `:core:mcp`
1. Remove `alias(libs.plugins.android.library)` from `build.gradle.kts`
2. Remove the `android { namespace = ... }` block
3. Remove `androidTarget()` from the `kotlin { }` block
4. Rename `src/commonMain/kotlin/` → `src/jvmMain/kotlin/` (use `git mv` to preserve blame)
5. Rename `src/commonTest/kotlin/` → `src/jvmTest/kotlin/` — merge into the existing `jvmTest` directory if one exists
6. Update `sourceSets { commonMain.dependencies { ... } }` → `sourceSets { jvmMain.dependencies { ... } }` in build.gradle.kts
7. Delete the (now empty) `androidMain` directory if it still exists
8. Verify `:app` and any other dependents still resolve — they pick up the `jvm()` target artifact
9. Verify `:core:mcp:jvmTest` still runs

### `:core:tunnel`
1. Move the three permissions from `core/tunnel/src/androidMain/AndroidManifest.xml` to `app/src/main/AndroidManifest.xml` (or `work/src/main/AndroidManifest.xml` — whichever module owns the runtime need). Verify no duplication.
2. Apply the same `:core:mcp` steps 1-9 above.
3. Move any contents of `src/androidMain/kotlin/` into `src/jvmMain/kotlin/` (e.g. `Frame.kt`).
4. Confirm `app-release.apk` still declares the same permission set at the end.

### `:core:bridge`
1. No plugin/target changes (already pure KMP with `jvm()` only).
2. Rename `src/commonMain/kotlin/` → `src/jvmMain/kotlin/`
3. Update `sourceSets { commonMain ... }` → `sourceSets { jvmMain ... }` in build.gradle.kts
4. Existing `src/jvmTest/` stays put.

## Tests
- `./gradlew assemble testDebugUnitTest` across all modules — all green
- Compare `AndroidManifest.xml` of the assembled APK before/after via `aapt dump permissions app-release.apk` — permission set must be identical
- Build-time comparison: record `./gradlew clean assemble` duration before and after, expect a measurable drop from dropping AGP overhead on two modules

## Scope guard
- Does NOT drop the KMP plugin. We're renaming source sets, not reclassifying modules.
- Does NOT add new abstractions. Does NOT split modules further. Pure plugin/target/layout cleanup.

Monkopedia:
## Summary

Fixes #129. Reshape `:core:mcp`, `:core:tunnel`, and `:core:bridge` so their build config and source layout match what the code actually is: JVM-only Kotlin libraries.

- `:core:mcp`: drop `com.android.library` plugin, drop `androidTarget()`, drop the `android {}` block. Rename `src/commonMain` to `src/jvmMain`. Collapse the `commonMain`/`commonTest` source sets into `jvmMain`/`jvmTest`.
- `:core:tunnel`: delete dead `src/androidMain/` (its manifest was the only thing contributing `INTERNET`/`WAKE_LOCK`/`FOREGROUND_SERVICE`; its Kotlin files were stale duplicates of `commonMain` that were never compiled). Move `WAKE_LOCK` to `:work`'s manifest alongside the existing `FOREGROUND_SERVICE*` declarations; `INTERNET` already lives on `:app`. Rename `src/commonMain`/`src/commonTest` to `src/jvmMain`/`src/jvmTest` (merge into existing). Collapse `CsrGenerator`'s redundant `expect`/`actual` split into a single concrete class (required now that both sides live in the same source set).
- `:core:bridge`: no plugin/target changes — already pure KMP `jvm()`. Just rename `src/commonMain` to `src/jvmMain`.
- Update the now-stale comment in `:app/src/main/AndroidManifest.xml` describing where `WAKE_LOCK`/`FOREGROUND_SERVICE*` come from.

The KMP plugin stays on every module so adding a future target is a plain addition rather than a restructure.

## Test plan

- [x] `:core:mcp:assemble :core:mcp:jvmTest` pass
- [x] `:core:tunnel:assemble :core:tunnel:jvmTest` pass
- [x] `:core:bridge:assemble :core:bridge:jvmTest` pass
- [x] `assemble testDebugUnitTest` pass across the whole project
- [x] `ktlintCheck` passes
- [x] `:app:assembleRelease` APK declares the same 50 permissions as before the change (verified with `aapt2 dump permissions`): `INTERNET`, `WAKE_LOCK`, `FOREGROUND_SERVICE`, `FOREGROUND_SERVICE_DATA_SYNC`, and the full Health Connect set are all present.


Monkopedia:
Superseded — cherry-picked directly to main (commits 8d10ceb6, 8073474b, b0c63cfb). We do not merge through PRs for agent work.

Issue #156, opened 2026-04-15
Migrate :integrations/notifications to McpTool DSL


Monkopedia:
Follow-up to #123 (McpTool DSL framework landed + :outreach migrated).

## Task
Convert `integrations/src/main/java/com/rousecontext/integrations/notifications/NotificationMcpProvider.kt` from hand-rolled `server.addTool(...)` calls to `McpTool` subclasses registered via `server.registerTool { ... }`.

Of note: `registerPerformAction` currently trips detekt's LongMethod rule and is a good candidate for splitting into a dedicated `PerformActionTool` subclass as part of this migration.

## Reference
- Framework: `core/mcp/src/jvmMain/kotlin/com/rousecontext/mcp/tool/`
- README: `core/mcp/src/jvmMain/kotlin/com/rousecontext/mcp/tool/README.md`
- Pattern example: `integrations/src/main/kotlin/com/rousecontext/integrations/outreach/OutreachMcpProvider.kt` (migrated in #123)

## Acceptance
- Wire format for success + tool-logic errors byte-identical to existing
- Existing notification provider tests pass unchanged (or get shorter)
- `./gradlew :integrations:testDebugUnitTest ktlintCheck detekt` all green
- LongMethod suppression on `registerPerformAction` can be dropped

Issue #155, opened 2026-04-15
Migrate :integrations/health to McpTool DSL


Monkopedia:
Follow-up to #123 (McpTool DSL framework landed + :outreach migrated).

## Task
Convert `integrations/src/main/java/com/rousecontext/integrations/health/HealthConnectMcpProvider.kt` (and any siblings in that package) from hand-rolled `server.addTool(...)` calls to `McpTool` subclasses registered via `server.registerTool { ... }`.

## Reference
- Framework: `core/mcp/src/jvmMain/kotlin/com/rousecontext/mcp/tool/`
- README: `core/mcp/src/jvmMain/kotlin/com/rousecontext/mcp/tool/README.md`
- Pattern example: `integrations/src/main/kotlin/com/rousecontext/integrations/outreach/OutreachMcpProvider.kt` (migrated in #123)

## Acceptance
- Wire format for success + tool-logic errors byte-identical to existing
- Existing health provider tests pass unchanged (or get shorter)
- `./gradlew :integrations:testDebugUnitTest ktlintCheck detekt` all green
- `InstantParam` is already available for ISO-8601 time-window arguments (since / until)

Issue #154, opened 2026-04-15
Migrate :integrations/usage to McpTool DSL


Monkopedia:
Follow-up to #123 (McpTool DSL framework landed + :outreach migrated).

## Task
Convert `integrations/src/main/kotlin/com/rousecontext/integrations/usage/UsageMcpProvider.kt` from hand-rolled `server.addTool(...)` calls to `McpTool` subclasses registered via `server.registerTool { ... }`.

## Reference
- Framework: `core/mcp/src/jvmMain/kotlin/com/rousecontext/mcp/tool/`
- README: `core/mcp/src/jvmMain/kotlin/com/rousecontext/mcp/tool/README.md`
- Pattern example: `integrations/src/main/kotlin/com/rousecontext/integrations/outreach/OutreachMcpProvider.kt` (migrated in #123)

## Tools to migrate
- `get_app_usage`
- `get_usage_summary`
- `get_usage_events`
- `compare_usage`

## Acceptance
- Wire format for success + tool-logic errors byte-identical to existing
- Existing `UsageMcpProviderTest` passes unchanged (or gets shorter)
- `./gradlew :integrations:testDebugUnitTest ktlintCheck detekt` all green

PR #157, opened 2026-04-15
Fix #123: McpTool DSL + migrate :integrations/outreach

### LINKED ISSUE #123: Clikt-style McpTool base class in :core:mcp to eliminate per-tool registration boilerplate
## Problem
Every MCP tool follows the same 15-25 line registration shape:

```kotlin
server.addTool(
    name = "launch_app",
    description = "Launch an installed app on the device by package name",
    inputSchema = launchAppSchema()   // hand-built JSON schema
) { request ->
    val packageName = request.params.arguments?.get("package_name")
        ?.jsonPrimitive?.content ?: return@addTool errorResult("Missing package_name")
    // ... actual logic ...
    successResult("...")
}
```

`outreach/src/main/kotlin/com/rousecontext/outreach/OutreachMcpProvider.kt` has **10** of these. Similar duplication in `usage/`, `notifications/`, `health/`.

Issues with the current shape:
- Hand-built JSON schemas drift from actual argument parsing (schema says `required`, handler forgets to check — silent bug).
- Param extraction boilerplate repeats literally: `?.get(k)?.jsonPrimitive?.content ?: errorResult(...)`.
- Description and name live in two places (schema builder + `addTool` call).
- Result wrapping (`successResult`, `errorResult`) is manual.
- Concurrency concerns (e.g. `withContext(Dispatchers.Main)`) live alongside argument parsing.

## Proposed design — `McpTool` base class (Clikt-inspired)

Location: `core/mcp/src/commonMain/kotlin/com/rousecontext/mcp/tool/` (so `:health`, `:outreach`, `:usage`, `:notifications` can all reuse).

```kotlin
abstract class McpTool {
    abstract val name: String
    abstract val description: String

    // Delegate-backed param declarations. Registered automatically.
    protected fun stringParam(name: String, description: String) = StringParam(name, description)
    protected fun intParam(name: String, description: String) = IntParam(name, description)
    protected fun boolParam(name: String, description: String) = BoolParam(name, description)
    protected fun <T : Enum<T>> enumParam(name: String, description: String, kClass: KClass<T>) = ...
    protected fun mapParam(name: String, description: String) = MapParam(name, description)
    // Modifiers: .optional(), .default(value), .choices("a", "b"), .range(0..100)

    abstract suspend fun execute(): ToolResult

    // Framework-internal: called by the registry during dispatch
    internal fun buildSchema(): ToolSchema = ...
    internal suspend fun invoke(args: JsonObject?): CallToolResult = ...
}

sealed class ToolResult {
    data class Success(val text: String, val meta: JsonObject? = null) : ToolResult()
    data class Error(val message: String) : ToolResult()
    data class Json(val value: JsonElement) : ToolResult()
}
```

### Core authoring pattern
```kotlin
class LaunchAppTool(
    private val context: Context,
    private val launchNotifier: LaunchRequestNotifierApi?,
) : McpTool() {
    override val name = "launch_app"
    override val description = "Launch an installed app by package name"

    val packageName by stringParam("package_name", "Package name, e.g. com.example.app")
    val extras by mapParam("extras", "String-valued intent extras").optional()

    override suspend fun execute(): ToolResult {
        val intent = context.packageManager.getLaunchIntentForPackage(packageName)
            ?: return ToolResult.Error("App not found: $packageName")
        extras?.forEach { (k, v) -> intent.putExtra(k, v) }
        // ...
        return ToolResult.Success("Launched $packageName")
    }
}

// Registration — one line per tool
server.registerTool { LaunchAppTool(context, launchNotifier) }
```

### Required/optional/default/choices must read naturally
```kotlin
val title    by stringParam("title", "Notification title")                  // required
val priority by intParam("priority", "Priority 0-5").default(3).range(0..5) // optional with default + range
val mode     by enumParam("mode", "DND mode", DndMode::class)               // required enum
val tags     by listParam<String>("tags", "Tag filter").optional()          // optional list
val since    by instantParam("since", "ISO-8601 start time").optional()     // optional Instant
```

### Non-goals for the DSL
- No hidden lifecycle magic. `execute()` is a plain `suspend fun`. No "before" / "after" hooks.
- No global registry. Tools are values, registered on a `Server` explicitly.
- No reflection-at-runtime beyond what's needed to walk property delegates during factory introspection.
- No annotation-processor / codegen step. Property delegates + runtime reflection only.

### Docs
- Land with a `core/mcp/src/commonMain/kotlin/com/rousecontext/mcp/tool/README.md` (or KDoc on `McpTool`) showing:
  - The 15-line "minimal tool" template
  - Every param type with one example each
  - Modifier chain examples (`.optional().default(...).choices(...)`)
  - Migration pattern: before/after for one real tool (`launch_app`) so providers know exactly what to change
- No essay. Readable top-to-bottom in two minutes.

## Param types needed (from audit of current providers)
- `string` — required/optional, default, choices
- `int` — required/optional, default, range
- `boolean` — required/optional, default
- `enum` — typed enum choice (Kotlin enum → JSON schema enum)
- `map<string, string>` — intent extras, header-like payloads
- `list<string>` — record-types lists, tag filters
- `Instant` (ISO-8601) — start/end times for usage, health queries

### Unit tests — framework internals
One dedicated test file per concern. At minimum:

**`ParamExtractionTest`** — for EACH param type:
- extracts a valid value correctly
- returns a clear error when required param is missing (`\"Missing required parameter 'foo'\"`)
- uses the default when optional param is missing
- rejects wrong JSON type with a specific message (`\"Parameter 'foo' must be a string, got number\"`)
- rejects values outside range / not in enum / not in choices list
- handles null JSON values (treat as missing)
- rejects unknown extra params strictly? or loosely? — decide and test the choice

**`SchemaGenerationTest`** — for EACH param type and combination:
- Emits correct JSON Schema (`type`, `description`, `enum`, `default`, `minimum`, `maximum`)
- `required` array matches non-optional params exactly
- Map and list types produce correct nested schemas
- Description text is forwarded verbatim
- Golden-file assertions: known tool shape → exact expected schema JSON (to catch accidental changes to emitted schema format)

**`ToolResultConversionTest`**:
- `Success` → `CallToolResult` with `isError == false`, content as `TextContent`
- `Error` → `CallToolResult` with `isError == true`, message preserved
- `Json` → `CallToolResult` with JSON-serialized TextContent
- Metadata field propagation

**`McpToolRegistrationTest`** — the factory + dispatcher layer:
- Factory is called once at registration (for schema) and once per invocation (for fresh instance, no state bleed)
- Two invocations with different args don't see each other's state
- Exception thrown in `execute()` maps to `ToolResult.Error` with the exception message (not unchecked propagation)
- Cancellation during `execute()` is honored (suspend test with `CoroutineScope.cancel()`)

### Integration tests — end-to-end through the real SDK Server
**`McpToolEndToEndTest`**:
- Register a tiny test tool built with the DSL on a real `Server`
- Invoke via `ClientConnection` (the SDK's test connection)
- Assert `tools/list` returns the generated schema correctly
- Assert `tools/call` with valid args returns the expected result
- Assert `tools/call` with missing/invalid args returns a proper MCP error response, not a crash

### Migration regression tests
After each provider migrates (e.g. `:outreach`), the provider's existing tests (`OutreachMcpProviderTest`) MUST:
- Pass without change beyond the factory construction
- Get **shorter**, not longer — any test that was asserting boilerplate behavior (missing-param error messages, etc.) should be deletable because the framework handles it
- Add a \"tool count and names match expected\" assertion per provider to catch drift

### Property-based tests (nice to have)
- Random JSON object → any param-extraction function should never throw (returns Error instead)
- Random param combinations → schema generation should always produce valid JSON Schema (optional kotlinx-serialization validation)

## Migration plan
1. **PR 1**: Framework + tests. No provider changes. Green CI.
2. **PR 2**: Migrate `:outreach` (most tools; best stress test). Tests should shrink.
3. **PR 3**: Migrate `:usage`.
4. **PR 4**: Migrate `:notifications` and `:health` (they have fewer tools).
5. **PR 5**: Delete the hand-built schema helpers (`launchAppSchema()`, `stringParamSchema(...)`, etc.) once nothing references them.

## Out of scope
- Output schemas. MCP's `outputSchema` is optional; adding later is straightforward.
- Streaming tool responses. Current tools are all request/response.
- Tool introspection for doc-gen. Nice-to-have, not required.

## Why not Koog or a third-party wrapper
- Koog brings the whole agent framework; we only want tool authoring.
- Third-party MCP wrappers are mostly single-person repos tracking an unstable protocol.
- In-repo keeps it ~200-300 LOC, we own it, tracks our needs, no external drift.

## Verify
`./gradlew :core:mcp:testDebugUnitTest :outreach:testDebugUnitTest ktlintCheck detekt` — all green at each migration step.

## Gate on #124
#124 consolidates `:outreach`, `:usage`, `:health`, `:notifications/capture` into one `:integrations` module. Land #124 first so this DSL migration happens in one module instead of four. Running this first then consolidating would duplicate effort and make #124's move-diff noisier.

Monkopedia:
## Summary
- Land the `McpTool` Clikt-style DSL in `:core:mcp` (framework + 5 test suites + 2-minute README)
- Migrate `:integrations/outreach` (ten tools) to the new shape as the first consumer; outreach provider shrinks from 883 → 547 lines
- Filed #154, #155, #156 for the three remaining provider migrations (usage / health / notifications)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

## Test plan
- [x] `./gradlew :core:mcp:jvmTest` — five new suites green (param extraction, schema generation, ToolResult conversion, registration/factory lifecycle, end-to-end through real Server)
- [x] `./gradlew :integrations:testDebugUnitTest` — existing `OutreachMcpProviderTest` passes unchanged against the migrated provider
- [x] `./gradlew :app:testDebugUnitTest` — no regressions
- [x] `./gradlew ktlintCheck detekt` — detekt sits at 72 weighted issues (baseline ~70; no new issues in the added framework / migrated code)

Framework: 628 LOC main + 698 LOC tests across seven test/helper files. Wire format for migrated tool successes and tool-logic errors is byte-identical; framework-level validation (missing required param, type mismatch, range, choices) is the only output path whose format changed and no existing tests / user-facing contracts exercised those paths.

Follow-ups (mechanical): #154, #155, #156.

### DIFF
--- a/core/mcp/src/jvmMain/kotlin/com/rousecontext/mcp/tool/McpTool.kt
+++ b/core/mcp/src/jvmMain/kotlin/com/rousecontext/mcp/tool/McpTool.kt
@@ -1,0 +1,139 @@
+package com.rousecontext.mcp.tool
+
+import com.rousecontext.mcp.tool.params.BoolParam
+import com.rousecontext.mcp.tool.params.EnumParam
+import com.rousecontext.mcp.tool.params.InstantParam
+import com.rousecontext.mcp.tool.params.IntParam
+import com.rousecontext.mcp.tool.params.ListParam
+import com.rousecontext.mcp.tool.params.MapParam
+import com.rousecontext.mcp.tool.params.ParamDef
+import com.rousecontext.mcp.tool.params.ParamExtract
+import com.rousecontext.mcp.tool.params.StringParam
+import io.modelcontextprotocol.kotlin.sdk.types.ToolSchema
+import kotlin.properties.ReadOnlyProperty
+import kotlin.reflect.KClass
+import kotlin.reflect.KProperty
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.buildJsonObject
+
+/**
+ * Base class for tools. A tool authored against [McpTool] declares its name,
+ * description, and typed parameters via delegate helpers, then overrides
+ * [execute] with the business logic.
+ *
+ * Instances are cheap — the framework creates a fresh one per call, so it is
+ * safe to hold request-scoped state as regular fields. Do not share instances
+ * across calls.
+ *
+ * See `core/mcp/src/jvmMain/kotlin/com/rousecontext/mcp/tool/README.md` for
+ * the full authoring guide.
+ */
+abstract class McpTool {
+    abstract val name: String
+    abstract val description: String
+
+    private val params = mutableListOf<ParamBinding<*>>()
+
+    /** Set by [invoke] before [execute] runs, so param delegates can read it. */
+    private var boundArgs: JsonObject? = null
+
+    /** Raw JSON arguments for tools that need to reach beyond the typed params. */
+    protected val rawArgs: JsonObject?
+        get() = boundArgs
+
+    // ---------- param declarations ----------
+
+    protected fun stringParam(name: String, description: String): StringParam =
+        StringParam(name, description)
+
+    protected fun intParam(name: String, description: String): IntParam =
+        IntParam(name, description)
+
+    protected fun boolParam(name: String, description: String): BoolParam =
+        BoolParam(name, description)
+
+    protected fun <T : Enum<T>> enumParam(
+        name: String,
+        description: String,
+        kClass: KClass<T>
+    ): EnumParam<T> = EnumParam(name, description, kClass)
+
+    protected fun mapParam(name: String, description: String): MapParam =
+        MapParam(name, description)
+
+    protected fun listParam(name: String, description: String): ListParam =
+        ListParam(name, description)
+
+    protected fun instantParam(name: String, description: String): InstantParam =
+        InstantParam(name, description)
+
+    // ---------- delegate wiring ----------
+
+    /**
+     * Binding of a [ParamDef] to a concrete property. Holds the most recent
+     * extracted value so [getValue] can return it.
+     */
+    private class ParamBinding<T>(val propertyName: String, val param: ParamDef<T>) :
+        ReadOnlyProperty<McpTool, T?> {
+        var cached: T? = null
+        override fun getValue(thisRef: McpTool, property: KProperty<*>): T? = cached
+    }
+
+    /**
+     * `val packageName by stringParam(...)` — captured here. We register the
+     * binding, then hand back a [ReadOnlyProperty] that reads the cached value.
+     */
+    protected operator fun <T, P : ParamDef<T>> P.provideDelegate(
+        thisRef: McpTool,
+        property: KProperty<*>
+    ): ReadOnlyProperty<McpTool, T?> {
+        val binding = ParamBinding(property.name, this)
+        thisRef.params.add(binding)
+        return binding
+    }
+
+    // ---------- framework-internal ----------
+
+    fun buildSchema(): ToolSchema {
+        val properties = buildJsonObject {
+            for (binding in params) {
+                put(binding.param.name, binding.param.schema())
+            }
+        }
+        val required = params
+            .filter { it.param.required }
+            .map { it.param.name }
+        return ToolSchema(properties = properties, required = required)
+    }
+
+    fun paramNames(): List<String> = params.map { it.param.name }
+
+    /**
+     * Extract all params, populate bindings, and run [execute]. Any exception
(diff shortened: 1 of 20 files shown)


Monkopedia:
Cherry-picked to main as 6cb8090d + ce2de387.

PR #452, opened 2026-05-26
Stop FGS gracefully on dataSync run-time timeout (#450)

monkopedia-coder:
## Summary
Fixes the `ForegroundServiceDidNotStopInTimeException` crashes (#450, and #451 closed as a dup).

Android 15+ enforces a cumulative ~6h/24h limit on `dataSync` foreground services. When the budget is exhausted **while the service is running**, the system calls `Service.onTimeout()` and gives the service a couple seconds to stop — otherwise it force-kills the process with `ForegroundServiceDidNotStopInTimeException`.

`TunnelForegroundService` already handled the symmetric **start-time** case (`ForegroundServiceStartNotAllowedException` → post `FgsLimitNotifier` + `stopSelf`), but never overrode `onTimeout`, so a budget that ran out **mid-session** crashed silently. v1.0.2's keep-Ktor-alive-across-idle change makes the service stay foregrounded longer, accelerating accumulation toward the cap — an accelerant, not the root cause (the missing handler predates it).

## Change
- Override `onTimeout(startId, fgsType)` to mirror the start-time path: set `intentionalDisconnect`, post the existing `FgsLimitNotifier`, and `stopSelf()`.
- Robolectric test (`@Config sdk=35`) asserting `onTimeout(dataSync)` stops the service, suppresses reconnect, and posts the notification.
- `docs/ux-decisions.md` entry (per `.claude/rules/ux-changes.md`) for the new lifecycle moment at which the budget notification can appear. Approved in-session by Jason 2026-05-26.

## Test plan
- [x] `:work:testDebugUnitTest` green (new test fails without the override, passes with it)
- [x] `ktlintCheck` + `detekt` clean
- [x] Built + launched on a Pixel 6 Pro running Android 17 (SDK 37) — clean startup, no crash, confirms the code loads/runs on the OS that enforces the timeout
- [ ] Real 6h `dataSync` cap **not** synthetically reproducible (no adb fast-forward for the cumulative budget); handler is unit-verified

Fixes #450

Issue #450, opened 2026-05-25
Crash: RemoteServiceException$ForegroundServiceDidNotStopInTimeException in FcmReceiver.startTunnelService


Monkopedia:
Crashlytics-Issue-ID: 3030a097e669f7d7ddd121fa77f35785

**Crashlytics link:** https://console.firebase.google.com/u/1/project/rouse-context/crashlytics/app/android:com.rousecontext/issues/3030a097e669f7d7ddd121fa77f35785

**App version(s) affected:** 1.0.2

**Device model / Android OS version:** OnePlus CPH2581 / Android 16

**Affected users:** 1 user / 2 events

**Stack trace:**
```
Fatal Exception: android.app.RemoteServiceException$ForegroundServiceDidNotStopInTimeException
A foreground service of type dataSync did not stop within its timeout: ComponentInfo{com.rousecontext/com.rousecontext.work.TunnelForegroundService}

android.app.ActivityThread.generateForegroundServiceDidNotStopInTimeException (ActivityThread.java:2653)
com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1074)

Caused by android.app.StackTrace
Last startServiceCommon() call for this service was made here

android.app.ContextImpl.startServiceCommon (ContextImpl.java:2095)
androidx.core.content.ContextCompat.startForegroundService (ContextCompat.java:563)
com.rousecontext.work.FcmReceiver.startTunnelService (FcmReceiver.java:96)
com.rousecontext.work.FcmReceiver.onMessageReceived$lambda$0 (FcmReceiver.java:39)
com.rousecontext.work.FcmReceiver.dispatchIfIntegrationsEnabled (FcmReceiver.kt:71)
com.rousecontext.work.FcmReceiver.onMessageReceived (FcmReceiver.kt:38)
com.google.firebase.messaging.FirebaseMessagingService.dispatchMessage (FirebaseMessagingService.java:254)
java.lang.Thread.run (Thread.java:1564)
```

Filed automatically from Crashlytics.

PR #456, opened 2026-06-02
Wire the dormant Connection settings cluster (#453)

### LINKED ISSUE #453: Idle-timeout Settings controls are non-functional (slider discards value, disable-switch unwired)
**Category:** test-gap / dead-code / inconsistency
**Where:**
- `app/src/main/java/com/rousecontext/app/ui/viewmodels/SettingsViewModel.kt:161` (`setIdleTimeout`)
- `app/src/main/java/com/rousecontext/app/ui/screens/SettingsScreen.kt:236-261` (idle-timeout dropdown + disable switch), `SettingsScreen.kt:139-140` (state fields)
- `app/src/main/java/com/rousecontext/app/ui/navigation/destinations/SettingsDestination.kt:77-92` (wiring)
- effective timeout lives in `app/src/main/java/com/rousecontext/app/di/AppModule.kt:128` (`IDLE_TIMEOUT_MS = 5 * 60 * 1000L`)

**Current:** The Settings screen renders an idle-timeout section with a "2 / 5 / 10 min" dropdown and a "Disable timeout" switch, but neither control changes anything:

```kotlin
fun setIdleTimeout(minutes: Int) {
    // TODO: persist to DataStore when idle timeout DataStore is added
    refresh()
}
```

`setIdleTimeout` discards its `minutes` argument and only calls `refresh()`. `SettingsUiState.idleTimeoutMinutes` is a hardcoded default (`5`) that nothing ever updates, so the dropdown always shows "5 min" regardless of selection. The "Disable timeout" switch is worse: `onDisableTimeoutToggled` is never wired in `SettingsDestinationContent` (it keeps its `{}` default), and `idleTimeoutDisabled` is a hardcoded `false`. The value actually used at runtime is the fixed `IDLE_TIMEOUT_MS` constant in `AppModule`. The only test (`SettingsViewModelTest` "initial state uses default settings") just asserts the hardcoded `5`, so the dead path is invisible to the suite.

**Proposed:** Pick one direction (this is why it's staged `needs-decision`, not auto-worked):
- **(a) Wire it up** — back idle-timeout minutes + disable-toggle with DataStore (mirroring the existing `SecurityCheckPreferences`/`CertRenewalPreferences` pattern), have `IdleTimeoutManager`/`AppModule` read that value instead of the constant, and add ViewModel tests that assert a selection round-trips. Note: this is a real UX behavior change (a previously-inert control becomes live), so it needs the `docs/ux-decisions.md` treatment per `.claude/rules/ux-changes.md`.
- **(b) Remove the controls** — drop the dropdown, the disable switch, and the dead `idleTimeoutMinutes`/`idleTimeoutDisabled` state + `setIdleTimeout` until the feature is built. Also UX-affecting (removes a visible control).

**Why:** A user can today select "10 min" or flip "Disable timeout" and nothing happens — a silent no-op control is a trust/correctness problem, and the dead state + unwired callback are a maintenance trap (they look load-bearing). Either resolution removes the misleading surface and closes a real test gap.
**Effort:** small refactor (b) / larger (a, includes DataStore + plumbing into :work + tests)

monkopedia-coder:
## Summary
Fixes #453. The whole "Connection" settings cluster was built as UI but never wired to the ViewModel/destination — four inert controls:

| Control | Was |
|---|---|
| Idle-timeout dropdown | `setIdleTimeout` discarded its arg; value hardcoded 5 |
| "Disable timeout" switch | callback unwired; gated on `batteryOptimizationExempt` hardcoded `false` |
| Battery-optimization warning | `showBatteryWarning` hardcoded `true` → shown to every user, always |
| "Fix this" button | callback unwired |

## Changes
- **Persist** idle-timeout minutes + disable flag in `AppStatePreferences` (mirrors the existing `securityCheckIntervalHours` int-pref).
- **`IdleTimeoutManager`** now takes a `suspend () -> Long?` provider, read each CONNECTED cycle (`null` = disabled → timer never arms), so a settings change takes effect on the next wake with no service restart. `AppModule` supplies it from `AppStatePreferences`; the old fixed `IDLE_TIMEOUT_MS` constant is removed.
- **Battery status**: `batteryOptimizationExempt` + `showBatteryWarning` now derive from the real `PowerManager` state via a new shared `BatteryOptimization` helper (lifted from `BugReportUriBuilder`, which now reuses it). Re-read on Settings `ON_RESUME`.
- **"Fix this"** opens `ACTION_IGNORE_BATTERY_OPTIMIZATION_SETTINGS` (no Play-restricted permission, unlike the direct-request intent).

## Tests
- `AppStatePreferencesTest`: idle-minutes + disable round-trip.
- `IdleTimeoutTest`: dynamic provider value, disabled-never-arms, value re-read per cycle.
- `BatteryOptimizationTest`: status read + intent action.
- `SettingsViewModelTest`: persist calls + state reflection for minutes/disable/battery.

## Verification
- [x] `:app` + `:work` `testDebugUnitTest`, `ktlintCheck`, `detekt` all green
- [x] Built + ran on a Pixel 6 Pro (Android 17): selecting **10 min** updates the dropdown live **and survives a force-stop + relaunch** (DataStore round-trip on cold start); Settings renders with no crash
- Note: the FCM-wake → `IdleTimeoutManager` path itself isn't exercisable in CI; the manager logic is unit-tested and the persist/read round-trip is device-verified

## UX
`docs/ux-decisions.md` entry added — the one rule-touching bit is activating the battery-opt **system screen** via the (already-present) "Fix this" button. Scope approved in-session.

Behavior note: with the idle timeout disabled the tunnel stays foregrounded until the 6h `dataSync` FGS cap stops it (the #452 path) — intended semantics.

Fixes #453

### DIFF
--- a/app/src/main/java/com/rousecontext/app/di/AppModule.kt
+++ b/app/src/main/java/com/rousecontext/app/di/AppModule.kt
@@ -38,6 +38,7 @@ import com.rousecontext.app.state.PreferencesSnapshotHolder
 import com.rousecontext.app.state.SecurityAlertGate
 import com.rousecontext.app.state.ThemePreference
 import com.rousecontext.app.state.notificationPermissionFlow
+import com.rousecontext.app.support.BatteryOptimization
 import com.rousecontext.app.support.BugReportUriBuilder
 import com.rousecontext.app.support.FirebaseCrashReporter
 import com.rousecontext.app.token.RoomTokenStore
@@ -123,9 +124,6 @@ import org.koin.core.module.dsl.viewModel
 import org.koin.core.qualifier.named
 import org.koin.dsl.bind
 import org.koin.dsl.module
-
-/** Idle timeout before disconnecting the tunnel after all streams close. */
-private const val IDLE_TIMEOUT_MS = 5 * 60 * 1000L
 
 /** How long to wait for graceful MCP session shutdown before forcing tunnel disconnect. */
 private const val GRACEFUL_SHUTDOWN_TIMEOUT_MS = 5_000L
@@ -585,8 +583,18 @@ val appModule = module {
         val tunnelClient: TunnelClient = get()
         val mcpSession: McpSession = get()
         val recorder: SpuriousWakeRecorder = get()
+        val appStatePreferences: AppStatePreferences = get()
         IdleTimeoutManager(
-            timeoutMillis = IDLE_TIMEOUT_MS,
+            // Read the user's idle-timeout setting fresh each wake cycle.
+            // Disabled => null => the timer never arms (the service then runs
+            // until the FGS dataSync budget or an FCM idle-stop intervenes).
+            timeoutProvider = {
+                if (appStatePreferences.idleTimeoutDisabled()) {
+                    null
+                } else {
+                    appStatePreferences.idleTimeoutMinutes() * 60_000L
+                }
+            },
             onTimeout = {
                 withTimeoutOrNull(GRACEFUL_SHUTDOWN_TIMEOUT_MS) { mcpSession.shutdown() }
                 tunnelClient.disconnect()
@@ -626,6 +634,7 @@ val appModule = module {
             integrations = get(),
             securityCheckPreferences = get(),
             appStatePreferences = get(),
+            batteryExemptProvider = { BatteryOptimization.isExempt(androidContext()) },
             spuriousWakesFlow = SettingsViewModel.spuriousWakeStatsFlow(get())
         )
     }
--- a/app/src/main/java/com/rousecontext/app/state/AppStatePreferences.kt
+++ b/app/src/main/java/com/rousecontext/app/state/AppStatePreferences.kt
@@ -37,6 +37,32 @@ class AppStatePreferences(private val context: Context) {
         }
     }
 
+    suspend fun idleTimeoutMinutes(): Int =
+        dataStore.data.first()[KEY_IDLE_TIMEOUT_MINUTES] ?: DEFAULT_IDLE_TIMEOUT_MINUTES
+
+    fun observeIdleTimeoutMinutes(): Flow<Int> = dataStore.data.map {
+        it[KEY_IDLE_TIMEOUT_MINUTES] ?: DEFAULT_IDLE_TIMEOUT_MINUTES
+    }
+
+    suspend fun setIdleTimeoutMinutes(value: Int) {
+        dataStore.edit { prefs ->
+            prefs[KEY_IDLE_TIMEOUT_MINUTES] = value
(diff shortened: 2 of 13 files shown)


monkopedia-reviewer (approved):
Reviewed as tier-1 bug fix (rouse-context policy, default tier for `bug_fix`).

**Wiring correctness — verified:**
- `setIdleTimeoutMinutes` / `setIdleTimeoutDisabled` write to DataStore and read back correctly; `AppStatePreferencesTest` covers defaults, write→read, and observable emissions for both keys.
- `timeoutProvider` lambda in `AppModule` calls the two `suspend fun` prefs reads inside the suspend lambda — correct. Arithmetic `minutes * 60_000L` is Int × Long = Long — correct.
- Disabled path: `timeoutProvider` returns `null` → `if (timeoutMillis != null)` is skipped → timer never arms. Correct.
- Battery-exempt status: `batteryExemptProvider()` called inside the `combine` lambda on every refresh trigger emission; `DisposableEffect` on `ON_RESUME` calls `viewModel.refresh()` so returning from the system settings screen re-reads the PowerManager state.
- "Fix this" uses `ACTION_IGNORE_BATTERY_OPTIMIZATION_SETTINGS` — no `REQUEST_IGNORE_BATTERY_OPTIMIZATIONS` permission needed; Play-policy-safe.

**Tests — non-tautological:**
- `disabled provider never arms the timer`: advances 60 min of virtual time, asserts `disconnected` not completed AND `timeoutFired` stays false.
- `provider is re-read each cycle`: mutates the captured var between cycles, asserts the second cycle honors the new value.
- `SettingsViewModelTest`: `coVerify { prefs.setIdleTimeoutMinutes(10) }` verifies the actual DataStore call; state-reflection tests use `awaitLoaded()` to drain the loading emission cleanly.
- `BatteryOptimizationTest`: Robolectric shadow flips `PowerManager` state — real behavior verified.

**#450/#451 teardown area:** The only structural change to `IdleTimeoutManager` adds a null-check around the existing `timerJob = launch { ... }` block. When disabled, no new coroutines are launched; when enabled, behavior is identical to the old fixed-constant path. `TunnelForegroundServiceLifecycleTest` updated correctly. No regression surface.

**Quality:** Idiomatic Kotlin throughout (destructuring, `Quint` slot reuse, `mockk(relaxed=true)` with explicit stubs). No dead code, no wildcard imports, no `GlobalScope`, structured concurrency consistent with the rest of the ViewModel layer. `startActivitySafely` vs `openUriSafely` are correctly separate helpers. UX decision log entry present and complete.

CI: Build & Test ✓, Lint ✓, Android CI test ✓ — all green.

Issue #414, opened 2026-04-26
Investigate: ~20s latency on first cold-start after fresh onboarding (vs ~3s on warm-cached cold-starts)


Monkopedia:
## Observation

A cold-start (process killed via \`cmd activity stop-app\`, then AI client connects, then process should respawn and serve the request) takes:

- **~20s** when run immediately after fresh onboarding (\`pm clear\` → onboard → e2e). Sometimes succeeds at-the-wire (one observed run: 21498ms with HTTP 200), sometimes loses the race against the relay's 20s wakeup timeout and EOFs (one observed run: 20226ms with TLS EOF).
- **~3s** on warm-cached cold-starts (run repeatedly without \`pm clear\`).

20s closely matches the relay's \`wait_for_device_wakeup\` budget. So the latency floor on fresh-onboarded state is "device tunnel reconnect after process kill" racing against that 20s ceiling.

## Why this isn't explained yet

Initial speculation pointed at "Firebase init / Koin bootstrap / WS handshake from cold" but those aren't 20-second operations on Android. Repeat cold-starts being fast (~3s) means the underlying restart-and-reconnect chain CAN finish quickly. Something specific about the **first cold-start after fresh onboarding** is taking 17-20 extra seconds vs subsequent cold-starts.

This is either a real product bug or a test-methodology artifact. We don't have a plausible explanation either way — investigation needed before declaring v0.1 ready.

### From relay (via gcloud + journalctl) on a reproduction at 2026-04-26 14:51:29 UTC

\`\`\`
T+0.000s: TCP connection arrives, decision=DevicePassthrough subdomain="stony"
T+0.005s: Secret matched from cache
T+0.005s: Relay opens TLS to firestore.googleapis.com  (look up device record / FCM token)
T+0.044s: Relay opens TLS to oauth2.googleapis.com    (FCM auth)
T+0.062s: Relay opens TLS to fcm.googleapis.com       (send wake)
T+20.226s: client EOF (wakeup-timeout fired)
\`\`\`

Relay successfully sent FCM wake. Then waited 20s for device's WS+mTLS session to register. Device session never came back in time.

Test command: \`cmd activity stop-app com.rousecontext.debug\` (the right primitive — see #394). New device PID was confirmed after the curl finished, so the process *did* eventually restart.

### From e2e test (PR #396 / ColdStartEndToEndTest) at 2026-04-24

Same flow, with auth token. One run completed at 21498ms (HTTP 200) — at-the-wire success.

[...]

PR #457, opened 2026-06-08
Extract shared ReadinessGate (#454)

### LINKED ISSUE #454: Extract the hand-rolled first-emission 'readiness gate' duplicated across 4 classes
**Category:** duplication
**Where:**
- `app/src/main/java/com/rousecontext/app/registry/IntegrationProviderRegistry.kt:50-99`
- `app/src/main/java/com/rousecontext/app/state/DeviceRegistrationStatus.kt:53-100`
- `app/src/main/java/com/rousecontext/app/state/SecurityAlertGate.kt:40-74`
- `app/src/main/java/com/rousecontext/app/registry/OutreachIntegration.kt` (`readyDeferred`/`readyLatch`/`awaitReady`/`awaitReadyBlocking`)

**Current:** Four classes independently hand-roll the identical "suspend/block until the backing flow has emitted at least once" primitive. Each declares the same pair:

```kotlin
private val readyDeferred = CompletableDeferred<Unit>()
private val readyLatch = CountDownLatch(1)

private fun signalReady() {
    readyDeferred.complete(Unit)
    if (readyLatch.count > 0) readyLatch.countDown()
}

suspend fun awaitReady() { readyDeferred.await() }
fun awaitReadyBlocking(timeoutMs: Long): Boolean =
    readyLatch.await(timeoutMs, TimeUnit.MILLISECONDS)
```

The bodies are byte-for-byte the same (comments included in two of them). `OutreachIntegration.awaitReadyBlocking` is even documented as "not currently exercised in production but provided for parity" — a copy-paste wart that exists only because the idiom was duplicated rather than shared. This gate is load-bearing on the cold-start FCM-wake path (#414/#416), so the duplicated correctness logic (idempotent signal, deferred + latch staying in sync) is exactly the kind of thing that should live in one tested place.

**Proposed:** Extract a small `ReadinessGate` (or `FirstEmissionGate`) helper — owns the `CompletableDeferred` + `CountDownLatch`, exposes `signalReady()` / `awaitReady()` / `awaitReadyBlocking(timeoutMs)` — and have all four classes delegate to it. Put a focused unit test on the helper (idempotent signal, blocking timeout returns false before signal / true after). Likely lives in `:core:mcp` or a small shared util alongside the `ProviderRegistry` interface that already defines this contract.

**Why:** One tested implementation of a concurrency primitive instead of four copies; removes drift risk (the four `signalReady` bodies must stay identical by hand today) and lets the unused `OutreachIntegration.awaitReadyBlocking` parity method disappear naturally.
**Effort:** small refactor

monkopedia-coder:
## What

Four classes independently hand-rolled the identical "suspend/block until the backing flow has emitted at least once" primitive — a `CompletableDeferred<Unit>` + `CountDownLatch(1)` pair with byte-for-byte identical `signalReady()` / `awaitReady()` / `awaitReadyBlocking(timeoutMs)` bodies. This change extracts a single tested `ReadinessGate` helper and has all four classes delegate to it.

## Where the helper lives + why

`core/mcp/src/jvmMain/kotlin/com/rousecontext/mcp/core/ReadinessGate.kt`, alongside the `ProviderRegistry` interface that already documents this exact `awaitReady` / `awaitReadyBlocking` contract (issue #414/#416). All four consumers are `:app`-only and `:app` already depends on `:core:mcp`, so this introduces no new dependency edge — it's the lowest-friction home that all four can import. `CompletableDeferred` is already used elsewhere in `:core:mcp/jvmMain`; `CountDownLatch` is JVM stdlib.

## Pure refactor — no behavior change

Each class's observable behavior is identical. The gate preserves the existing semantics exactly: idempotent signal, deferred + latch kept in sync, blocking variant timing out via `CountDownLatch.await(timeoutMs, MILLISECONDS)`. When `signalReady` fires and the `signalReady` bodies are unchanged in timing/placement — only the storage of the primitive moved. All pre-existing tests (`IntegrationProviderRegistryTest`, `OutreachIntegrationTest`, `SecurityAlertGateTest`, `DeviceRegistrationStatusTest`) pass unmodified.

## Test added

`core/mcp/src/jvmTest/.../ReadinessGateTest.kt` — covers: `awaitReadyBlocking` false before signal / true after; `awaitReady` suspends until signalled then resumes; `awaitReady` returns immediately when already signalled; double/triple signal is safe and idempotent.

## Note on the OutreachIntegration "parity" method

The issue suggested `OutreachIntegration.awaitReadyBlocking` ("not currently exercised in production but provided for parity") could disappear. It does NOT fall away naturally — the existing `OutreachIntegrationTest` directly exercises `integration.awaitReadyBlocking(...)`. Removing it would require modifying an existing test, which the refactor rules forbid. It's kept and now delegates to the gate (one-liner, no longer a copy-pasted body).

## Migrated
- `IntegrationProviderRegistry`
- `DeviceRegistrationStatus`
- `SecurityAlertGate`
- `OutreachIntegration`

## Verification
- `:core:mcp:jvmTest` + `:app:testDebugUnitTest` — pass
- `ktlintCheck` + `detekt` — clean (ktlintFormat applied to collapse the now-shorter delegating one-liners)

Fixes #454

### DIFF
--- a/app/src/main/java/com/rousecontext/app/registry/IntegrationProviderRegistry.kt
+++ b/app/src/main/java/com/rousecontext/app/registry/IntegrationProviderRegistry.kt
@@ -4,9 +4,7 @@ import com.rousecontext.api.IntegrationStateStore
 import com.rousecontext.api.McpIntegration
 import com.rousecontext.mcp.core.McpServerProvider
 import com.rousecontext.mcp.core.ProviderRegistry
-import java.util.concurrent.CountDownLatch
-import java.util.concurrent.TimeUnit
-import kotlinx.coroutines.CompletableDeferred
+import com.rousecontext.mcp.core.ReadinessGate
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
@@ -47,8 +45,7 @@ class IntegrationProviderRegistry(
     /** Live snapshot of enabled integration IDs. Exposed for tests and UI. */
     val enabledIds: StateFlow<Set<String>> = _enabledIds.asStateFlow()
 
-    private val readyDeferred = CompletableDeferred<Unit>()
-    private val readyLatch = CountDownLatch(1)
+    private val readinessGate = ReadinessGate()
 
     init {
         val flows = integrations.map { integration ->
@@ -73,11 +70,8 @@ class IntegrationProviderRegistry(
     }
 
     private fun signalReady() {
-        // Idempotent: subsequent emissions just no-op on the already-completed signals.
-        readyDeferred.complete(Unit)
-        if (readyLatch.count > 0) {
-            readyLatch.countDown()
-        }
+        // Idempotent: subsequent emissions just no-op on the already-ready gate.
+        readinessGate.signalReady()
     }
 
     override fun providerForPath(path: String): McpServerProvider? {
@@ -92,9 +86,9 @@ class IntegrationProviderRegistry(
         .toSet()
 
     override suspend fun awaitReady() {
-        readyDeferred.await()
+        readinessGate.awaitReady()
     }
 
     override fun awaitReadyBlocking(timeoutMs: Long): Boolean =
-        readyLatch.await(timeoutMs, TimeUnit.MILLISECONDS)
+        readinessGate.awaitReadyBlocking(timeoutMs)
 }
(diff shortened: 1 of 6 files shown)


monkopedia-reviewer (approved):
Tier-1 vetted (equivalence confirmed across all 4 sites, CI green, tests sound, no quality issues). Visibility escalation resolved: rouse-context is an app, not a library — :core:mcp has no external consumers, so ReadinessGate being public is internal architecture/tech-debt, not a published API contract. Jason signed off in-session that this isn't an API decision. Merging.

Issue #446, opened 2026-05-03
Claude Code connections die on phone sleep — need graceful MCP session shutdown


Monkopedia:
## Problem

Claude Code MCP connections reach the relay, get spliced to the device via mux stream, but the TLS handshake hangs indefinitely. Claude web connections through the same relay/device work fine.

## Evidence

Relay logs show:
- Claude web (160.79.106.35) — streams complete in ~500ms
- Claude Code (71.184.193.67) — streams hang for ~5 minutes then get connection reset

Phone logs show "New mux stream N, starting session handler" but nothing after — the TLS accept in SessionHandler.handleStream hangs.

Direct curl through the relay works. openssl s_client works. Only Claude Code's TLS client fails.

## Hypothesis

TLS version or cipher mismatch between Claude Code's client and Android's SSLEngine, causing a silent handshake hang.

## Next steps

- Add detailed logging around TLS accept to capture what's happening
- Check if Claude Code sends a different ClientHello than curl/web

Monkopedia:
## Deep investigation results

### What we confirmed works
- TLS handshake completes successfully (TLSv1.3, ~200ms) for connections from both Claude web and Claude Code
- Manual MCP calls using Claude Code's exact OAuth token succeed via curl (initialize + tools/call)
- The server responds correctly to all HTTP methods (POST/GET/DELETE /mcp)
- ALPN was added (advertising http/1.1) — doesn't change the behavior

### What's actually happening

Claude Code has **two independent code paths**:

1. **Session/connection establishment** — works. Establishes TCP to relay, TLS handshake completes on the phone. This is what triggers the phone going active.
2. **Tool call dispatch** — broken. Returns instantly with `Error POSTing to endpoint:` (empty URL). Never sends HTTP bytes.

Evidence:
- `ss` shows Claude Code establishes a fresh TCP connection to the relay on each attempt (different local port each time)
- The relay splices it through as a new mux stream, TLS completes on the phone
- But the tool call returns instantly before any HTTP data could round-trip
- The connection then sits idle until it eventually resets

### Network flow confirmed
```
kaladin (192.168.68.92) → Claude proxy (71.184.193.67) → relay (35.209.161.222) → phone mux → phone TLS accept → OK
```
But the actual POST with the JSON-RPC tool call never arrives.

### Conclusion
This is a Claude Code client bug. The MCP client's tool call dispatch has a broken or empty endpoint URL in memory, even though:
- The credentials file has the correct URL
[...]

PR #458, opened 2026-06-08
Deduplicate graceful-shutdown sequence (#455)

### LINKED ISSUE #455: Deduplicate the graceful-shutdown sequence (mcpSession.shutdown + tunnel.disconnect) and its 5s timeout constant
**Category:** duplication
**Where:**
- `work/src/main/kotlin/com/rousecontext/work/TunnelForegroundService.kt:297-300` (`onDestroy`) and `:357` (`GRACEFUL_SHUTDOWN_TIMEOUT_MS = 5_000L`)
- `app/src/main/java/com/rousecontext/app/di/AppModule.kt:590-593` (`IdleTimeoutManager.onTimeout`) and `:131` (`GRACEFUL_SHUTDOWN_TIMEOUT_MS = 5_000L`)

**Current:** The same graceful-teardown sequence introduced for #446 appears in two places:

```kotlin
// AppModule.kt — IdleTimeoutManager onTimeout
withTimeoutOrNull(GRACEFUL_SHUTDOWN_TIMEOUT_MS) { mcpSession.shutdown() }
tunnelClient.disconnect()
```
```kotlin
// TunnelForegroundService.onDestroy
withTimeoutOrNull(GRACEFUL_SHUTDOWN_TIMEOUT_MS) { mcpSession.shutdown() }
tunnelClient.disconnect()
```

The body is identical and the `GRACEFUL_SHUTDOWN_TIMEOUT_MS = 5_000L` constant is declared separately in each module. Both encode the same #446 contract ("close MCP sessions cleanly, bounded by a timeout, then drop the tunnel"). If one is ever tuned (timeout value, ordering, adding a step) the other silently drifts — and they live in different modules, so the divergence wouldn't be obvious in review.

**Proposed:** Factor the sequence into one named operation — e.g. a `suspend fun gracefulTunnelShutdown(mcpSession, tunnelClient)` (or a method on a small teardown helper) with a single `GRACEFUL_SHUTDOWN_TIMEOUT_MS`. Both the `onTimeout` lambda and `onDestroy` call it. Keep the constant in one home (`:work` or `:core`) and reference it from both. This also gives the #446 contract a single unit-testable seam.

**Why:** Single source of truth for a correctness-sensitive teardown path; removes the cross-module copy and the duplicated magic constant so the two callers can't drift.
**Effort:** small refactor

monkopedia-coder:
## Summary

The #446 graceful-teardown sequence was copied into two modules, each with its own `GRACEFUL_SHUTDOWN_TIMEOUT_MS = 5_000L`:

- `TunnelForegroundService.onDestroy` (`:work`)
- the `IdleTimeoutManager.onTimeout` lambda wired in `AppModule` (`:app`), with a top-level private constant

Both encode the same contract — "close MCP sessions cleanly, bounded by a timeout, then drop the tunnel." Tuning one (timeout, ordering, an added step) would silently drift the other, and they live in different modules.

## Change

Factored the sequence into a single shared seam:

```kotlin
suspend fun gracefulTunnelShutdown(mcpSession: McpSession, tunnelClient: TunnelClient) {
    withTimeoutOrNull(GRACEFUL_SHUTDOWN_TIMEOUT_MS) { mcpSession.shutdown() }
    tunnelClient.disconnect()
}
```

**Where it lives:** new `work/src/main/kotlin/com/rousecontext/work/GracefulTunnelShutdown.kt`, holding both the function and the single `GRACEFUL_SHUTDOWN_TIMEOUT_MS`. `:work` is the lowest module that already depends on both `:core:mcp` (`McpSession`) and `:core:tunnel` (`TunnelClient`), and `:app` depends on `:work` — so both call sites import it cleanly with no dependency-direction change. Confirmed it compiles in both modules.

Both call sites now invoke the shared function; the two duplicate constant declarations are removed.

## Pure refactor — no behavior change

- Ordering preserved: shutdown-bounded-by-timeout **then** disconnect.
- The 5s timeout value is unchanged.
- `onDestroy` keeps its exact `lifecycleScope.launch { ... }` structure (the body is just replaced by the call); scoping/launch behavior untouched.
- The function introduces no scope of its own — it's a plain `suspend fun` called from the existing scopes.

## Test

Added `GracefulTunnelShutdownTest` covering the #446 contract directly:
- calls `mcpSession.shutdown()` then `tunnelClient.disconnect()` in order, and
- a hanging `shutdown()` is bounded by the timeout while `disconnect()` still runs.

The existing `TunnelForegroundServiceLifecycleTest` (onDestroy/onTimeout coverage) passes unmodified.

## Verification

- `:work:testDebugUnitTest` — green
- `:app:testDebugUnitTest` — green
- `ktlintCheck detekt` — green

Fixes #455

### DIFF
--- a/app/src/main/java/com/rousecontext/app/di/AppModule.kt
+++ b/app/src/main/java/com/rousecontext/app/di/AppModule.kt
@@ -114,19 +114,16 @@ import com.rousecontext.work.SpuriousWakePreferences
 import com.rousecontext.work.SpuriousWakeRecorder
 import com.rousecontext.work.StoredCertVerifierSource
 import com.rousecontext.work.WakelockManager
+import com.rousecontext.work.gracefulTunnelShutdown
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.withTimeoutOrNull
 import org.koin.android.ext.koin.androidContext
 import org.koin.core.module.dsl.singleOf
 import org.koin.core.module.dsl.viewModel
 import org.koin.core.qualifier.named
 import org.koin.dsl.bind
 import org.koin.dsl.module
-
-/** How long to wait for graceful MCP session shutdown before forcing tunnel disconnect. */
-private const val GRACEFUL_SHUTDOWN_TIMEOUT_MS = 5_000L
 
 /**
  * crt.sh issuer names that are legitimate for Rouse Context certificates.
@@ -596,8 +593,7 @@ val appModule = module {
                 }
             },
             onTimeout = {
-                withTimeoutOrNull(GRACEFUL_SHUTDOWN_TIMEOUT_MS) { mcpSession.shutdown() }
-                tunnelClient.disconnect()
+                gracefulTunnelShutdown(mcpSession, tunnelClient)
             },
             recorder = recorder
         )
--- a/work/src/main/kotlin/com/rousecontext/work/GracefulTunnelShutdown.kt
+++ b/work/src/main/kotlin/com/rousecontext/work/GracefulTunnelShutdown.kt
@@ -1,0 +1,26 @@
+package com.rousecontext.work
+
+import com.rousecontext.mcp.core.McpSession
+import com.rousecontext.tunnel.TunnelClient
+import kotlinx.coroutines.withTimeoutOrNull
+
+/**
+ * How long to wait for graceful MCP session shutdown before forcing tunnel
+ * disconnect. Single source of truth for the #446 teardown contract.
+ */
+const val GRACEFUL_SHUTDOWN_TIMEOUT_MS = 5_000L
+
+/**
+ * Tears down the tunnel gracefully per the #446 contract: close MCP sessions
+ * cleanly (bounded by [GRACEFUL_SHUTDOWN_TIMEOUT_MS] so a wedged transport
+ * can't block teardown forever), then drop the tunnel.
+ *
+ * This is the single shared seam called from both the
+ * [TunnelForegroundService] `onDestroy` path and the `IdleTimeoutManager`
+ * `onTimeout` lambda (wired in the :app Koin module). It deliberately does NOT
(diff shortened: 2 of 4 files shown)


monkopedia-reviewer (approved):
Verified the dedup. Summary of what I checked:

**Ordering equivalence — confirmed.** Both pre-refactor inlined sequences were byte-for-byte identical: `withTimeoutOrNull(5_000L) { mcpSession.shutdown() }` then `tunnelClient.disconnect()`. The extracted `gracefulTunnelShutdown` preserves that order exactly. Timeout value (5 s) unchanged.

**Scoping — clean.** `onDestroy` retains its `lifecycleScope.launch { gracefulTunnelShutdown(...) }` wrapper; no new scope is introduced. `onTimeout` is already called from within `IdleTimeoutManager`'s `coroutineScope`-backed `launch { }` — the function is a plain `suspend fun`, no scope capture.

**Double-shutdown — pre-existing, not regressed.** The `onTimeout → disconnect → DISCONNECTED → stopSelf() → onDestroy` path could run the sequence twice before and after this PR. The refactor doesn't change that. Not a regression.

**Module placement — correct.** `:work` already depends on both `:core:mcp` (`McpSession`) and `:core:tunnel` (`TunnelClient`), and `:app` depends on `:work`. No dependency-direction changes.

**Tests — meaningful.** `GracefulTunnelShutdownTest` asserts ordering with `coVerifyOrder` and separately asserts that a hanging `shutdown()` is bounded by the timeout while `disconnect()` still runs. Both use `CompletableDeferred` per the coroutine testing conventions. The existing `TunnelForegroundServiceLifecycleTest` is unmodified and passes. CI (test + lint + build) all green.

**One nit:** `GRACEFUL_SHUTDOWN_TIMEOUT_MS` is declared `public const val` but is only referenced within `:work` (the shared function + its test). The function itself must be `public` for cross-module use from `:app`, but the constant is an implementation detail and could be `internal` — tests in the same module see `internal`. Not blocking (CI passes, nothing outside `:work` uses it), but worth a follow-up cleanup.

Issue #256, opened 2026-04-18
HttpCtLogFetcher: retry with backoff on transient 5xx instead of surfacing warning on first failure


Monkopedia:
## Symptom

User got a 'CT log check: Could not reach CT log service: crt.sh returned HTTP 502' notification. crt.sh routinely 502s under load — it's a shared community service backed by Sectigo. A single failure is nearly always transient.

## Context

#227 correctly reclassifies this as "Could not reach" (not "Malformed"), which it is. But the `SecurityCheckWorker` surfaces every `Warning` as a user notification, so a single crt.sh hiccup = user notification.

## Proposed fix

In `core/tunnel/src/jvmMain/kotlin/com/rousecontext/tunnel/HttpCtLogFetcher.kt`: retry on 5xx HTTP status with exponential backoff. 3 attempts, 1s / 3s / 9s (or similar). 4xx stays a hard failure (query shape bug). If all retries fail, throw as today — `CtLogMonitor` classifies as Warning.

Keep the final timeout bounded: total should be ≤ 30s so the security-check worker doesn't hang forever.

## Alternative / complementary

In `SecurityCheckWorker` / notifier: only fire a user notification if a check returns `Warning` **N consecutive times** (e.g. 2 or 3). Transient warnings that self-heal on the next scheduled run never surface. More conservative but adds state.

Recommend **both**: retries handle the common-case 5xx flap, debouncing handles persistent-but-intermittent outages and network hiccups.

## Test

- `HttpCtLogFetcherTest`: fetcher called against a test server that returns 502 twice then 200 — assert fetch succeeds without error.
- `SecurityCheckWorkerTest` (if debouncing lands): consecutive warning from same source, assert notification posts only after threshold.

## Impact

User-facing notification noise. Not a correctness issue. Low severity but quality-of-life for release.

PR #459, opened 2026-06-08
Add Certspotter fallback for CT-log checks

monkopedia-coder:
## Problem

The Certificate Transparency security check (`CtLogMonitor`) queries crt.sh via `HttpCtLogFetcher`. crt.sh (community-run, maintained by Sectigo) frequently returns 502/503 for minutes at a time. The existing per-request retry (1s/3s/9s, #256) plus cross-run debounce (3 consecutive failures, #256/#429) reduce but don't eliminate user-facing "Could not reach CT log service" warnings, because crt.sh outages outlast the retry window. Those residual warnings are crt.sh being down, not a real reachability problem.

There is no GitHub issue for this — it's a direct user-requested reliability fix.

## Fix

Add a second, independent CT source (Certspotter) consulted **only after** crt.sh has already failed, so a crt.sh outage becomes a silent successful check instead of a warning — while preserving real security coverage.

- **`CompositeCtLogFetcher(primary, fallback)`** — `fetch` tries crt.sh; on any failure tries certspotter; if **both** fail, rethrows the **primary** (crt.sh) error so the monitor's "Could not reach CT log service: \<crt.sh message\>" wording stays accurate. `CancellationException` is rethrown explicitly (structured-concurrency rule).
- **`CertspotterCtLogFetcher`** — GET `api.certspotter.com/v1/issuances` (`domain`, `include_subdomains=true`, `expand=issuer`), then translates the certspotter issuance array into crt.sh-shaped JSON (`List<CtLogEntry>`) that `CtLogMonitor` already parses. Non-2xx throws `IOException`; network errors propagate; no internal retry (the composite only reaches it after crt.sh failed, staying within the 30s worker budget).
- Wire `AppModule`'s `CtLogFetcher` to `CompositeCtLogFetcher(primary = HttpCtLogFetcher(), fallback = CertspotterCtLogFetcher())`.

## Key correctness point (security path)

Certspotter's `issuer.name` string is **byte-identical** to crt.sh's `issuer_name` — same `C=US, O=..., CN=...` DN order and comma-space separators — verified against live data. So each issuance's `issuer.name` maps **straight** into the issuer field the monitor compares, with **no normalization** and **zero false-Alert risk**. A test parses the translated JSON with the same `Json` the monitor uses and asserts the issuer set is exactly the certspotter `issuer.name` values.

## Unchanged (no UX/error-surface change)

`CtLogMonitor` (classification), `HttpCtLogFetcher`, `SecurityCheckWorker`, the debounce, and the notification/warning copy are all untouched. This only makes the underlying fetch succeed more often, so there's no UX-surface change and no `docs/ux-decisions.md` entry is needed.

## Tests (new)

- `CertspotterCtLogFetcherTest`: canned certspotter response → translated JSON's parsed `issuer_name`s equal the exact certspotter `issuer.name` strings; empty array → `[]` (empty list); 502 → `IOException`.
- `CompositeCtLogFetcherTest`: primary success → fallback not invoked; primary fails → fallback used; both fail → primary's exception (message) rethrown.

## Verification

- `:core:tunnel:jvmTest` — green (existing `CtLogMonitorTest` + `HttpCtLogFetcherTest` pass unmodified)
- `:app:testDebugUnitTest` — green
- `ktlintCheck` + `detekt` — clean

### DIFF
--- a/app/src/main/java/com/rousecontext/app/di/AppModule.kt
+++ b/app/src/main/java/com/rousecontext/app/di/AppModule.kt
@@ -83,6 +83,8 @@ import com.rousecontext.notifications.create
 import com.rousecontext.tunnel.CertProvisioningFlow
 import com.rousecontext.tunnel.CertRenewalFlow
 import com.rousecontext.tunnel.CertificateStore
+import com.rousecontext.tunnel.CertspotterCtLogFetcher
+import com.rousecontext.tunnel.CompositeCtLogFetcher
 import com.rousecontext.tunnel.CsrGenerator
 import com.rousecontext.tunnel.CtLogFetcher
 import com.rousecontext.tunnel.CtLogMonitor
@@ -414,7 +416,16 @@ val appModule = module {
     }
 
     // --- Security checks (SecurityCheckWorker dependencies) ---
-    single<CtLogFetcher> { HttpCtLogFetcher() }
+    // crt.sh (primary) flaps with 502/503 for minutes at a time, outlasting
+    // HttpCtLogFetcher's retry window. Certspotter is queried as a fallback so a
+    // crt.sh outage becomes a silent successful check instead of a false
+    // "Could not reach CT log service" warning. See CompositeCtLogFetcher.
+    single<CtLogFetcher> {
+        CompositeCtLogFetcher(
+            primary = HttpCtLogFetcher(),
+            fallback = CertspotterCtLogFetcher()
+        )
+    }
     single {
         CtLogMonitor(
             certificateStore = get(),
--- a/core/tunnel/src/jvmMain/kotlin/com/rousecontext/tunnel/CertspotterCtLogFetcher.kt
+++ b/core/tunnel/src/jvmMain/kotlin/com/rousecontext/tunnel/CertspotterCtLogFetcher.kt
@@ -1,0 +1,66 @@
+package com.rousecontext.tunnel
+
+import io.ktor.client.HttpClient
+import io.ktor.client.request.get
+import io.ktor.client.request.parameter
+import io.ktor.client.statement.bodyAsText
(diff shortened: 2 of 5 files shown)


monkopedia-reviewer (approved):
Reviewed the full diff. All four correctness requirements verified:

**Issuer-format equivalence (the load-bearing one):** `CertspotterCtLogFetcher` maps `issuer.name` → `CtLogEntry(issuerName = it.issuer.name)` with zero normalization. The test locks this in with exact string equality (`C=US, O=Google Trust Services, CN=WR1` / `C=US, O=Let's Encrypt, CN=R13`), parsed via the same `Json { ignoreUnknownKeys = true }` the monitor uses. A DN-format drift from certspotter would break the assertion. Zero false-Alert risk confirmed.

**Query scope:** `domain=<subdomain>.<base>&include_subdomains=true&expand=issuer` — covers the wildcard `*.<device>.rousecontext.com` cert via `include_subdomains`. Correct.

**Empty array → Verified:** decodes `[]` → empty list → encodes back → monitor returns `Verified`. Test asserts `entries.isEmpty()`. ✓

**CompositeCtLogFetcher:** primary-success (fallback not called), primary-fail (fallback used), both-fail (primary exception rethrown, message verified exactly), CancellationException rethrown at both catch levels before the generic catch — all four tested or structurally confirmed. ✓

**Non-2xx → IOException:** status check before `bodyAsText()` — no chance of an HTML error body reaching the JSON parser. ✓

**Quality:** consistent with `HttpCtLogFetcher` in package, naming, `RelayApiClient.createDefaultClient()`, `ignoreUnknownKeys`, error-message idiom. Idiomatic Kotlin. No dead code, no wildcard imports. CI green (Lint + Build & Test + Android test all SUCCESS).

Issue #460, opened 2026-06-11
Epic: fully-FOSS build flavor for F-Droid (UnifiedPush + keypair auth + ACRA)


Monkopedia:
Tracking issue for shipping a **fully-FOSS build of the Android app** suitable for F-Droid, while keeping the existing Google/Play build unchanged.

## Decision: support both, don't go FOSS-only

We ship **two product flavors** off the same codebase:

- **`google`** — the current build, unchanged. FCM wake, Firebase anonymous auth, Crashlytics. Distributed as a sideloaded signed APK (we have no Play listing). Existing installs are unaffected — no migration, no distributor-app burden.
- **`foss`** — zero Firebase/Play-Services dependencies linked. UnifiedPush wake, device-keypair identity, ACRA crash reporting. Built by F-Droid from source; the signed APK also published on GitHub Releases.

The **relay supports both** auth + push paths simultaneously (a per-device discriminator), so one backend serves both flavors. Going FOSS-only is rejected: it would force every existing install onto a UnifiedPush distributor app — a UX regression for the current users, and pointless churn since the `google` build already works for them.

> **Note:** F-Droid audits the *APK only*. The relay can keep using a Google service account, Firestore, and FCM (for `google`-flavor devices) indefinitely. The only requirement is that the `foss` APK link no proprietary Google libraries.

## What's already in our favor

- **The keypair primitive already exists and is already primary renewal auth.** `relay/src/api/renew.rs:123-125` always verifies a CSR signature against the device's stored public key; Firebase is only an *optional* extra on the expired-cert path. The device already generates a hardware-backed ECDSA P-256 keypair (`DeviceKeyManager`) and registers its public key (`relay/src/api/register.rs:382`).
- **Firebase Auth is only truly load-bearing at initial `/register`** (`relay/src/api/register.rs:113-123`), where the Firebase token is the sole authenticator and the UID is the Firestore foreign key. That's the one place needing a new auth primitive — and the user has confirmed the design: register a public key, prove possession by signing a nonce/timestamp (JWT-style). We don't reuse identity across installs, so a fresh keypair == a fresh device (same semantics as Firebase anonymous on reinstall).
- **Crashlytics is already behind the `CrashReporter` interface** with a `NoOp` variant — drop-in swap per flavor.
[...]

Issue #463, opened 2026-06-11
FOSS #3: UnifiedPush wake path (foss app + relay), dual-routed with FCM


Monkopedia:
Part of #460. Depends on #461 (flavor foundation — must land first). **UX-gated** — the distributor-install onboarding step needs explicit approval + a `docs/ux-decisions.md` entry before implementation starts (`.claude/rules/ux-changes.md`).

## Goal

Wake `foss` devices via [UnifiedPush](https://unifiedpush.org/) instead of FCM. Relay supports **both** FCM (google) and UnifiedPush (foss) per device.

## Background (seams already present)

- Relay has an `FcmClient` trait (`relay/src/fcm.rs:31`) — add a sibling sender, don't rewrite.
- Android already funnels both wake + renew nudges through a flavor-agnostic dispatcher, `FcmDispatch.resolve()` (`work/.../FcmDispatch.kt`). The `foss` receiver reuses it.
- Wake payload is `{"type":"wake"}`, renew is `{"type":"renew"}` (`relay/src/fcm.rs:42-53`).
- Device reports its token over the tunnel WebSocket: `{"type":"fcm_token","token":...}` (`core/tunnel/.../TunnelClientImpl.kt:227-231`).

## Work

**App (`foss` flavor):**
- Add the UnifiedPush connector library; implement a `MessagingReceiver` that, on push, calls the existing `FcmDispatch.resolve()` → `startTunnelService()` / cert-renewal enqueue. No `FirebaseMessagingService` in this flavor.
- On endpoint registration/refresh, report the **UnifiedPush endpoint URL** to the relay (generalize the WS message to carry a push-kind + endpoint; see relay change below). Mirror `FcmTokenRegistrar` behavior (push when connected, defer to next wake when not).
- First-run: a single **picker screen** ("Background delivery") lists installed UnifiedPush distributors as one-tap rows; when **no** distributor is installed it suggests ntfy + an "install another app" row that deep-links to F-Droid/Play. **ntfy is suggested only in the zero-installed case** — if any supported distributor is present (even a non-ntfy one), the list shows exactly the installed distributors + "install another app", never an "install ntfy" row. No multi-step install→connect→continue wizard. **This onboarding addition is the UX-gated part — get approval + decision-log entry first** (mocks rendered for review via `FossDeliveryMockTest`).
[...]

PR #468, opened 2026-06-13
Keypair device identity for foss flavor (#462)

### LINKED ISSUE #462: FOSS #2: keypair device identity — Firebase-free register + renew (relay supports both)
Part of #460. Depends on #461 (flavor foundation — must land first). **Touches the security model — security-sensitive review required.**

## Goal

Let a `foss` device register and renew certs with **no Firebase**, using its existing hardware-backed keypair to prove identity. The relay must accept **both** Firebase-token (google) and keypair (foss) auth.

## Background (already built)

- Device generates a hardware-backed ECDSA P-256 keypair: `core/tunnel/.../DeviceKeyManager.kt`.
- Renewal already verifies a CSR signature against the stored public key on **every** `/renew` (`relay/src/api/renew.rs:123-125`); Firebase is only an *optional* extra on the expired-cert path (`renew.rs:83-109`).
- Public key is registered in round-2 of registration today (`relay/src/api/register.rs:382`).
- Firebase is load-bearing **only** at initial `/register` (`relay/src/api/register.rs:113-123`) where the UID is the Firestore foreign key.

## Work

**Relay:**
- Extend `DeviceRecord` so device identity can be keyed on a **public-key thumbprint** (SHA-256 of the SPKI DER) as an alternative to `firebase_uid`. Add a Firestore lookup by thumbprint mirroring `find_device_by_uid`.
- `POST /register`: accept a keypair-auth variant — request carries the device public key + a signature over a server-issued or client-timestamped nonce (replay-protected: bound timestamp window, reject stale). Verify the signature with the supplied key, derive the thumbprint, generate + assign the subdomain keyed on the thumbprint. Keep the Firebase path unchanged.
- Expired-cert renewal (`/renew` Path B): accept a **keypair-signed timestamp token** in lieu of `firebase_token`, verified against the device's stored public key. The CSR signature is still always checked. Keep Firebase Path B working for google.
- Reject obviously bad input (malformed key, expired timestamp, thumbprint/record mismatch). Add unit tests for verify-success, bad-signature, stale-timestamp, and identity-mismatch.

**App (`foss` flavor):**
- Implement the `foss` device-auth binding: on first run, generate/load the keypair via `DeviceKeyManager`, build the registration proof (signed nonce/timestamp), call the keypair `/register`. No `FirebaseAnonymousAuthClient`.
- Implement the `foss` expired-renewal credentials provider: sign a fresh timestamp token instead of fetching a Firebase ID token (replaces `FirebaseRenewalAuthProvider` for this flavor).
- `google` flavor keeps the Firebase impls.

## Acceptance

- A `foss` build registers, provisions a cert, and renews (both valid-cert and expired-cert paths) end-to-end against the relay with no Firebase token in any request.
- `google` registration/renewal unchanged.
- Relay unit tests cover the new keypair verification paths; replay/stale-timestamp rejection tested.
- `cargo test && cargo clippy --all-targets --all-features -- -D warnings` clean; app `ktlintCheck detekt` clean.

## Design note

We don't reuse identity across installs (confirmed): a fresh keypair == a fresh device, matching Firebase-anonymous reinstall semantics. The nonce-signing scheme and replay window are a security decision — flag the exact scheme in the PR for review before merge.

monkopedia-coder:
Part of #460. Fixes #462. **Security-sensitive** (device registration/renewal auth) — requesting explicit user review (tier-3, no auto-merge).

## What this does

Replaces Firebase Anonymous Auth with a hardware-backed device-keypair identity for the **`foss`** flavor. The relay supports **both** auth paths simultaneously (a per-device discriminator), so one backend serves `google` (Firebase) and `foss` (keypair) devices. The `google` flavor's Firebase impls are untouched.

The keypair primitive already existed (`DeviceKeyManager`, ECDSA P-256, hardware-backed) and is already the primary renewal auth (CSR signature on every `/renew`). The only place Firebase was load-bearing was initial `/register` and the expired-cert `/renew` Path B. Those now have a keypair alternative.

## Exact keypair scheme (please vet this)

**What is signed.** The device signs the canonical byte string

```
rouse-context-keypair-auth:v1\n<purpose>\n<timestamp_secs>\n<nonce>
```

- `purpose` is `register` or `renew` — binds a proof to its endpoint so a register proof can't be replayed as a renew proof.
- `timestamp_secs` — device Unix time (decimal ASCII).
- `nonce` — base64 of 16 random bytes; makes each proof unique.

Signature = `SHA256withECDSA` (ECDSA-P256 over SHA-256), DER-encoded, base64 for transport — the identical mechanism already used for CSR-DER signatures. The canonical layout is mirrored byte-for-byte in `relay/src/keypair_auth.rs` and `core/tunnel/.../KeypairAuth.kt` (guarded by a contract test).

**Replay window.** The relay rejects any proof whose timestamp is more than **±300s** from its own clock (`REPLAY_WINDOW_SECS`). Stateless: no server-issued nonce, no persistence (the relay is memory-constrained). This is acceptable because the proof is *never the sole gate that mints a cert* — `/renew` **always** also verifies the CSR signature over the freshly-generated CSR, and a replay yields certs only for the device's own key (useless without the private key, which never leaves the Keystore).

**Identity / thumbprint.** A keypair device's stable identity is `thumbprint = base64(SHA-256(SPKI DER))` — the role the Firebase UID plays for `google`. Stored as `DeviceRecord.key_thumbprint` (Firestore foreign key; `find_device_by_thumbprint` mirrors `find_device_by_uid`). A fresh keypair == a fresh device, matching Firebase-anonymous reinstall semantics (no cross-install identity reuse).

**Per endpoint:**
- `POST /register` (round 1, keypair): carries `public_key` + `auth_timestamp`/`auth_nonce`/`auth_signature`. Relay verifies the proof with the *supplied* key (purpose `register`), derives the thumbprint, generates+assigns the subdomain keyed on the thumbprint. `firebase_uid` empty; `public_key` left empty (bound in round 2). The Firebase-only subdomain *reservation* flow is skipped for foss.
- `POST /register/certs` (round 2, keypair): **no token**. The relay derives the thumbprint from the **CSR's** public key and looks up the device by it. The CSR self-signature (already verified) proves possession; the thumbprint match proves identity, so no separate proof is needed. Binds `public_key`.
- `POST /renew` Path B (keypair): instead of `firebase_token`, carries a fresh `renew`-purpose proof verified against the device's **stored** public key. The CSR signature is still always checked. Firebase Path B is unchanged for `google`.

The same Keystore key (`rouse_device_key`) backs the proof signature, the CSR, and the registered public key, so all thumbprints/signatures are consistent.

## App (foss only)
- `KeypairDeviceCredentialProvider` (new) builds the signed registration proof from `DeviceKeyManager` + the Keystore signer.
- `KeypairRenewalAuthProvider` (new): `signCsr` for the valid-cert path; `acquireKeypairRenewalCredentials` (CSR sig + renew proof) for the expired path; returns `null` from `acquireFirebaseCredentials`.
- The three foss stub auth bindings (`NoOpAnonymousAuthClient`, `NoOpDeviceAuthTokenProvider`, `NoOpRenewalAuthProvider`) are removed.
- Shared flows/ViewModels now thread a flavor-agnostic `DeviceCredential` (sealed: `Firebase` or `Keypair`); the `google` Firebase path is behaviorally identical (backward-compat overloads retained).

## Tests
- Relay unit (`keypair_auth`): verify-success, bad-signature, signature-from-other-key, stale future/past timestamp, window-edge accept, purpose-mismatch, unknown-purpose, malformed-signature, invalid-key, thumbprint stability.
- Relay HTTP (`api_register_test`, `api_renew_test`): keypair register keyed on thumbprint, bad-sig 403, no-auth 400, round-2 unknown-thumbprint 404, full register→certs e2e bind; renew keypair valid proof, stale 403, forged 403. Firebase paths unchanged.
- Kotlin: `RelayApiClient` keypair wire-format test + canonical-message contract test.
- `cargo test` + `cargo clippy --all-targets --all-features -- -D warnings` clean; `:app:assembleFossDebug :app:assembleGoogleDebug :app:testGoogleDebugUnitTest`, `:work:testDebugUnitTest`, `:core:tunnel:jvmTest`, `ktlintCheck`, `detekt` all green.

## Notes for the reviewer
- **Replay window choice (300s, stateless)** is the core security decision; rationale above. Open to a server-issued single-use nonce if preferred over statelessness.
- **No foss DI verification test runs in CI** — the foss Koin graph is validated by `assembleFossDebug` (compile) only; every foss binding was hand-checked. The epic's on-device gate exercises it at runtime.
- **App-side real-relay integration test (`:app:integrationTest`) for the keypair flow was not added** to bound scope/risk; relay-side HTTP tests already exercise register+renew keypair end-to-end at the relay boundary, and the Kotlin client is unit-covered.


monkopedia-reviewer (commented):
I reviewed this PR to merge standard across both security and quality axes. Here's what I verified and what needs your judgment.

## What I reviewed and found sound

**keypair_auth.rs — core cryptography**
- `verify_proof` is correct: purpose is checked first (rejects unknown purposes immediately), then timestamp window (±300 s, enforced symmetrically on both future and past), then key parsing, then sig decode, then signature verification. Order is sensible.
- Domain-separation prefix (`rouse-context-keypair-auth:v1`) + `purpose` field are both baked into the signed canonical message, so a `register` proof produces different signed bytes than a `renew` proof — cross-purpose replay is structurally prevented, not just policy-blocked.
- Empty/malformed inputs are rejected with distinct error variants (`InvalidPublicKey`, `MalformedSignature`, `UnknownPurpose`, `StaleTimestamp`, `SignatureMismatch`), all surfaced as 400 or 403 as appropriate.
- Test coverage is comprehensive: verify-success, bad-sig, sig-from-other-key, stale future/past timestamp, edge-accept at window boundary, purpose-mismatch, unknown-purpose, malformed-sig, invalid-key, thumbprint stability.

**Replay window on both endpoints**
- `/register` — `resolve_register_identity` calls `verify_proof` with `now_unix_secs()`, ±300 s enforced.
- `/renew` — keypair block calls `verify_proof` against server `now_secs`, same 300 s window. Verified in `api_renew_test.rs`: stale-403 test uses `now_secs() - 3600`, well outside the window.
- The CSR signature (`signature` field) is unconditionally required and verified **after** the keypair block in `handle_renew`. If an attacker replays a valid keypair proof with a different CSR, the CSR signature check stops them — the relay would need both a valid proof AND a CSR signed by the private key to issue a cert. The private key never leaves the Keystore, so a replayed proof alone is useless.

**Thumbprint identity**
- `thumbprint = base64(SHA-256(SPKI DER))` computed in `keypair_auth::thumbprint`, used consistently in both round-1 (from received `public_key` field) and round-2 (derived fresh from CSR, via `extract_public_key_from_csr_der`). Round-2 identity depends on the CSR using the same key as round-1 — it will, since `DeviceKeyManager.getOrCreateKeyPair()` returns a stable Keystore key.
- `find_device_by_thumbprint` queries the `key_thumbprint` Firestore field; `find_device_by_uid` queries `firebase_uid`. The two identity spaces are disjoint — a keypair thumbprint cannot collide with or shadow a `firebase_uid` record.
- `google` devices always have `key_thumbprint = None`; Firestore `EQUAL` filter on a string value won't match null fields. No cross-flavor identity confusion.

**Firebase / google path regression**
- `DeviceCredential.Firebase` → `when (credential)` and `when (identity)` branches are explicit and unchanged in behavior.
- `FirebaseDeviceCredentialProvider` wraps existing `AnonymousAuthClient` / `DeviceAuthTokenProvider` without modification; `google` DI module adds it alongside the existing bindings.
- `renewExpired` tries keypair first (via `keypairProvider.obtain`), falls through to Firebase if `null` is returned. For `google`, `acquireKeypairRenewalCredentials` returns `null` (the interface default), so the Firebase path is taken as before.

**foss Koin graph**
- All three `NoOp*` stubs removed, replaced with `KeypairDeviceCredentialProvider` and `KeypairRenewalAuthProvider`, which receive their deps via `get()`. The dependencies (`DeviceKeyManager`, `DeviceKeystoreSigner`) are bound in `appModule`. Coherent at compile level; `assembleFossDebug` verifies this.

**CI**
- Build & Test ✓, Lint ✓, Rust/format ✓, integration suite (21 min) ✓ — all green on HEAD `73de12cf`.

**Minor non-blocking observations (not requesting changes)**
- `CertRenewer.renewWithFirebase` now calls `flow.renewExpired` internally and handles both keypair and Firebase. The method name is stale — a future cleanup (perhaps during #463/#464) would be a natural time to rename it `renewExpired` throughout.
- `registerCertsWithKeypair` has a `signature: String? = null` parameter that is never passed a non-null value from any call site. Effectively unused on the keypair path. Not wrong, but dead surface.
- In `handle_register_certs`, `extract_public_key_from_csr_der` is called twice for the keypair path (once to derive the thumbprint for the lookup, once at the step-4 re-verify). The code comment explains this is intentional to keep step-4 uniform across both paths, which is correct.

---

## Residual items for @Monkopedia — these need your judgment

The three items the PR description already flags are real open questions that warrant your call:

1. **±300 s stateless replay window.** A valid keypair proof can be replayed against the same device for up to 5 minutes after the device signed it. The PR argues this is acceptable because the CSR signature is an independent gate (a replayed proof gets you nothing without the private key). If you're concerned about the window, the alternative is a server-issued single-use nonce (requires a round-trip and some server state). No wrong answer — just make sure you've consciously chosen stateless-with-window over stateful-nonce.

2. **foss DI graph is compile-only verified.** `assembleFossDebug` checks the Koin graph compiles, but there's no CI job that starts the foss app and exercises the registration flow. The epic's on-device gate is the intended runtime check. Whether that's sufficient before merging is your call.

3. **No `:app:integrationTest` for the keypair path.** The relay HTTP tests in `api_register_test.rs` and `api_renew_test.rs` cover the register→certs e2e and renew at the relay boundary. The Kotlin `RelayApiClientTest` covers the wire format and canonical-message contract. There is no full-stack test through the app layer to the real relay for the keypair path. The PR bounds scope intentionally; this is a known gap, not an oversight.

Everything the bot can judge is sound. The three items above are the only reason this isn't ready to auto-merge.

Issue #368, opened 2026-04-22
Audit screen shows cached empty state after notification tap; deep-link extras also dropped


Monkopedia:
## Symptom

User tapped a per-tool-call notification, was deep-linked to the audit history screen, and the screen showed the empty state — despite the tool call having just happened (the notification is proof it was persisted by `RoomAuditListener`).

Ruled out by user reporting:
- **Not a filter issue.** The tool call was recent, and the default `DateFilterOption.TODAY` window would include it. If it were filter-related, scrolling through filter chips would reveal the entry — it did not.
- The state looks *cached*: the screen rendered the `groups.isEmpty()` empty-state UI, not the loading spinner, not a populated list.

## Suspected mechanics

`AuditHistoryViewModel.state` uses:

```kotlin
.stateIn(
    scope = viewModelScope,
    started = SharingStarted.WhileSubscribed(STOP_TIMEOUT_MS = 5_000L),
    initialValue = AuditHistoryState(isLoading = true)
)
```

`WhileSubscribed(5s)` keeps the upstream alive for 5 seconds after the last subscriber leaves. If the user:
1. Visits the audit screen (empty), causing the Flow to emit `groups = []`
2. Navigates away (e.g., home)
3. Makes a tool call — the notifier fires via `RoomAuditListener`, which inserts a row into the audit DB
4. Taps the notification within ~5s of step 2

...then step 4's subscription re-joins the still-alive upstream Flow. The last-emitted value was `[]`. Whether Room's `observeByDateRange` fires a new emission depends on a few race conditions:

- Did the `RoomAuditListener` insert complete before the UI re-subscribed?
- Is Room's invalidation fired on the main DB connection (vs. the one serving `observeByDateRange`)?
- Does `flatMapLatest` tear down and rebuild the inner Room flow on upstream restart — and if so, does the restart happen at the right moment?

There are a few failure shapes here, and the cleanest diagnosis is a Robolectric test that reproduces the sequence.

## Secondary bug (known, track separately if this agent doesn't subsume it)

`MainActivity.destinationForNotificationIntent` returns bare `Routes.AUDIT_BASE` (just `"audit"`) and never encodes the notification's intent extras into the nav route. That drops `scrollToCallId` / `startMillis` / `endMillis` entirely, so:
- Per-tool-call taps never scroll to the called row (they just land on audit)
- Summary taps never apply the session time window — they land on the TODAY filter

[...]

PR #471, opened 2026-06-13
Fix leaked Main dispatcher flake (#376)

### LINKED ISSUE #376: Flaky: TestMainDispatcher race in AuditHistoryViewModelRoomTest + OnboardingViewModelTest
Two unrelated test classes flake with what looks like a `TestMainDispatcher` race. Both surface once in local full-suite runs, pass on rerun and on CI. First observed during PR #375's agent pass.

## Suspects

- `app/src/test/java/com/rousecontext/app/ui/viewmodels/AuditHistoryViewModelRoomTest.kt` (added in #369)
- `app/src/test/java/com/rousecontext/app/ui/viewmodels/OnboardingViewModelTest.kt` (older)

## Symptom

Exact stack trace not captured; agent just noted the failures. Expected to be one of:
- `kotlinx.coroutines.test.UncompletedCoroutinesError`
- `Dispatchers.setMain` / `resetMain` ordering
- `TestScope` / `runTest` with a shared state flow that leaks between tests
- `Main` dispatcher set by a prior test not being torn down before the next class runs

## Action

- Next time either test fails, grab the full stack from the build report (`app/build/reports/tests/testDebugUnitTest/index.html`) and update this issue
- Audit setup/teardown in both test classes — confirm `Dispatchers.setMain` / `resetMain` is always paired, and `TestScope` is properly closed
- If the race is in a shared `StateFlow` or `WhileSubscribed` subscriber leaking across tests, consider injecting a per-test `CoroutineScope` that's explicitly cancelled in `@After`

## Non-goals

- Don't speculate-patch without a real stack trace. Single-observation flakes aren't enough signal to blindly rewrite tests.

## Priority

Low. Both passed on rerun and CI is green. Promote when a third observation hits.

## Related

- #341 — similar "flaky-on-first-run" pattern fixed by replacing polling with `CompletableDeferred`
- #263 — older low-priority flake on a similar "single observation, rerun-passes" profile

monkopedia-coder:
## The flake (grounded in CI)

CI run **27476494784**, "Build & Test / Unit tests":

- **Attempt 1 (`:app`):**
  - `AuditHistoryViewModelRoomTest.kt:83` → `java.lang.IllegalStateException`, `Caused by: java.lang.Throwable at TestMainDispatcher.kt:76`. Line 83 is `Dispatchers.resetMain()`.
  - `OnboardingViewModelTest > startOnboarding handles auth exception gracefully` → `IllegalStateException at TestMainDispatcher.kt:72`, `Caused by: Throwable at TestMainDispatcher.kt:86`.
- **Attempt 2 was `BUILD SUCCESSFUL`.** The `SecurityCheckSchedulerTest.setUp:45` line cited earlier was a benign `CloseGuard` SQLite-leak *warning* logged via `System.logW` during a passing run — not a failure. No `:work` test calls `setMain`, so there is no Main-dispatcher leak in `:work`. This change is therefore scoped to `:app`.

## Mechanism: a leaked Main override, single JVM

`TestMainDispatcher.kt:72/76/86` is `kotlinx.coroutines.test`'s internal `NonConcurrentlyModifiable` guard. Its setter throws `IllegalStateException("Dispatchers.Main is used concurrently with setting it")` when a coroutine has *read* Main (recorded location = the `Caused by` Throwable) while another `setMain`/`resetMain` writes it. Because unit tests run in a **single JVM** (`testOptions.unitTests` sets no `forkEvery`), one test leaving a coroutine dispatching on Main pollutes a **random subsequent** test's `setMain` — hence moving-target victims and no single-class local repro (the two April audits found nothing locally).

## The leaking call site (confirmed)

`AuditHistoryViewModelRoomTest`:
- pointed Main at the **shared `Dispatchers.IO`** pool (threads outlive the test), and
- tore down with a fire-and-forget `testScope.coroutineContext[Job]?.cancel()` (**no join**) before `resetMain()`, and
- **never cancelled the VM's `viewModelScope`** at all.

The VM's `stateIn` upstream / Room flow collector runs on `viewModelScope` (= `Dispatchers.Main.immediate` = IO here). After the test body it could still be dispatching on Main when `resetMain()` ran — which is exactly the attempt-1 self-failure at `:83`, and which leaves a coroutine reading Main past the reset to poison the next class's `setMain` (the `OnboardingViewModelTest` victim).

## Fix (structural, deterministic)

- **`MainDispatcherRule`** (new): pairs `setMain`/`resetMain` in a `TestRule` so the reset always runs in a `finally` around the whole test — `setMain` can never be left dangling by an early return/throw.
- **Dedicated single-thread executor as Main** (not shared IO): preserves the real, thread-backed cross-thread scheduler Room's invalidation tracker needs (an `UnconfinedTestDispatcher` would trampoline emissions and miss wakeups), while making the dispatcher fully **drainable** because the test owns it.
- **VM owned in a `ViewModelStore`**; `tearDown`:
  1. `viewModelStore.clear()` → cancels `viewModelScope` (stops the `stateIn` upstream + Room collectors),
  2. `mainExecutor.shutdown()` + `awaitTermination(...)` → blocks until the single worker thread is **idle and dead**,
  3. then the rule's `resetMain()` fires.
- Warmup collector joined with `cancelAndJoin` on the `runBlocking` scope.

## Why this is correct (not just "ran green N times")

The flake is a data race between a coroutine reading Main and `resetMain` writing it. The fix **removes the race condition itself**: after `awaitTermination` returns, the only thread that could dispatch on the test's Main dispatcher has terminated, so when `resetMain()` runs there is provably no concurrent reader of Main, and no coroutine can outlive the reset into the next test. Using a single-thread executor we own (vs. the shared, never-terminating IO pool) is what makes that termination observable and deterministic. The repeated-green runs below corroborate; the argument above is the proof.

## Verification

- `:app:testGoogleDebugUnitTest --rerun-tasks` full suite: **7 consecutive green** runs (5 + 2 after `ktlintFormat`).
- `AuditHistoryViewModelRoomTest` alone: green (the #368 regression assertion still holds).
- `:work:testDebugUnitTest --rerun-tasks`: 2 green (unaffected, confirmed no `setMain`).
- `ktlintCheck detekt`: clean.

Fixes #376.

### DIFF
--- a/app/src/test/java/com/rousecontext/app/testing/MainDispatcherRule.kt
+++ b/app/src/test/java/com/rousecontext/app/testing/MainDispatcherRule.kt
@@ -1,0 +1,48 @@
+package com.rousecontext.app.testing
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import kotlinx.coroutines.test.resetMain
+import kotlinx.coroutines.test.setMain
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+
+/**
+ * JUnit rule that installs [dispatcher] as `Dispatchers.Main` for the duration
+ * of a test and unconditionally restores the original Main afterwards.
+ *
+ * Pairing `setMain`/`resetMain` in a rule (rather than ad-hoc `@Before`/`@After`
+ * methods) makes the reset bulletproof: the rule's teardown runs in a `finally`
+ * around the whole test even if the body throws, so the Main dispatcher can
+ * never be left overridden and leak into a subsequent test in the same JVM.
+ *
+ * Unit tests run in a single JVM (`testOptions.unitTests` sets no `forkEvery`),
+ * so a leaked Main override surfaces as an `IllegalStateException` from
+ * `kotlinx.coroutines.test`'s internal `TestMainDispatcher` in a *random later*
+ * test's `setMain` — the moving-target flake tracked in issue #376.
+ *
+ * The rule's teardown only resets Main; it does NOT join coroutines. A test that
+ * routes Main to a *real* (thread-backed) dispatcher must ensure its own
+ * coroutines are cancelled and drained to quiescence in its `@After` (which runs
+ * before this rule's teardown) so nothing is still dispatching on Main when the
+ * reset fires.
+ */
+@OptIn(ExperimentalCoroutinesApi::class)
+class MainDispatcherRule(private val dispatcher: CoroutineDispatcher = UnconfinedTestDispatcher()) :
+    TestRule {
+
+    override fun apply(base: Statement, description: Description): Statement =
+        object : Statement() {
+            override fun evaluate() {
+                Dispatchers.setMain(dispatcher)
+                try {
+                    base.evaluate()
+                } finally {
+                    Dispatchers.resetMain()
+                }
+            }
+        }
+}
--- a/app/src/test/java/com/rousecontext/app/ui/viewmodels/AuditHistoryViewModelRoomTest.kt
+++ b/app/src/test/java/com/rousecontext/app/ui/viewmodels/AuditHistoryViewModelRoomTest.kt
@@ -1,29 +1,30 @@
 package com.rousecontext.app.ui.viewmodels
 
 import android.content.Context
+import androidx.lifecycle.ViewModelStore
 import androidx.room.Room
 import androidx.test.core.app.ApplicationProvider
+import com.rousecontext.app.testing.MainDispatcherRule
 import com.rousecontext.notifications.audit.AuditDatabase
 import com.rousecontext.notifications.audit.AuditEntry
+import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
Reviewed at tier-1 standard. The fix is structurally correct and the quality bar is met on both axes.

**Correctness:** The teardown ordering genuinely eliminates the race — `viewModelStore.clear()` cancels `viewModelScope` (queuing cancellation continuations onto `mainExecutor`), then `shutdown()` + `awaitTermination()` drains those continuations and kills the executor thread, then the `MainDispatcherRule`'s `finally` resets Main. After `awaitTermination` the thread is provably dead; there is no concurrent reader of Main when `resetMain()` fires, so the race condition from #376 is removed by construction, not by luck.

**`MainDispatcherRule`** is the standard `kotlinx-coroutines-test` idiom (setMain before evaluate, resetMain in finally). Well-placed in the `testing` package, clear KDoc explaining the flake mechanism.

**Scope discipline:** Only two test files in `:app`. PR body correctly explains the `:work` attempt-2 line was a benign CloseGuard warning, not a `setMain` leak.

**Test assertions:** The #368 regression assertion (re-subscribe after insert sees new row) is unchanged. `warmupJob` now launches on `this` (the `runBlocking` scope) instead of the old unscoped `testScope` — a structured-concurrency improvement per project rules.

**Quality:** No dead code, no over-verbosity, idiomatic Kotlin (`check()` for the awaitTermination guard, `DRAIN_TIMEOUT_MS` constant with a good comment). The `ViewModelStore.put()` teardown pattern matches `OnboardingViewModelTest`'s established usage.

CI green on all three checks. Auto-merging.

Issue #469, opened 2026-06-13
Close the foss keypair integration-test gap: foss-variant integration harness + keypair round-trip test


monkopedia-coder:
Follow-up from #462/#468 (keypair device identity), to close **immediately** (per Jason).

## The gap
#462 added the keypair register/renew auth path for the `foss` flavor, but there is **no `:app:integrationTest` coverage of the foss keypair flow against the real relay**. Today the keypair path is covered only at:
- the **relay HTTP boundary** (Rust tests: register→certs + renew keypair end-to-end), and
- the **Kotlin client unit** level (proof generation + a byte-level wire-format contract test).

What's missing is the **live round-trip**: the foss Android client actually talking to a running relay binary with keypair auth and getting/renewing a cert.

## Why it wasn't done in #462
The existing `:app:integrationTest` task reuses `testGoogleDebugUnitTest`'s classpath (set up in #461), so the harness runs under the **`google` variant** and loads the Firebase DI bindings. The keypair impls live in `app/src/foss/`, off that classpath — so exercising them end-to-end needs new harness infrastructure, which was scoped out of the security PR.

## Work
- Add a **foss-variant integration path** — e.g. a `fossIntegrationTest` Gradle task mirroring `integrationTest` but reusing `testFossDebugUnitTest`'s classpath (so the foss `DistributionModule` keypair providers are live). Adapt if a cleaner approach exists; note the chosen approach in the PR.
- Add a keypair **round-trip integration test** against the real relay binary (`cd relay && cargo build --features test-mode`): foss device registers via keypair proof → cert provisioning → renew via keypair → assert certs issued and the relay accepts the proofs. Mirror the existing google integration tests (e.g. `CertRotationIntegrationTest`, `ColdStartViaFcmTest`) for shape.
- Wire the new task into CI (`.github/workflows/android-ci.yml`) so it runs alongside the existing google `integrationTest`.

## Why this is worth building now
Beyond closing the #462 gap, the foss-variant integration harness is **reused by #463 (UnifiedPush) and #464 (ACRA)** — both have foss-flavor runtime paths that will want the same end-to-end coverage. Build it once here.

Part of #460.

PR #472, opened 2026-06-13
Relay: per-device push target + UnifiedPush wake/renew routing (#463 relay half)

monkopedia-coder:
Relay half of #463 (UnifiedPush wake path), split per the implementation-approach comment. Additive: **all existing FCM behavior is unchanged**; this only adds a second transport behind a per-device discriminator.

## Push-target model (additive)
`DeviceRecord` gains two `#[serde(default)]` fields:
- `push_kind: PushKind` (`fcm` | `unifiedpush`), defaulting to `Fcm`
- `push_endpoint: String` (the UnifiedPush POST URL; empty for FCM devices, which keep using `fcm_token`)

Legacy/`google` Firestore documents (no `push_kind`/`push_endpoint`) deserialize as `Fcm` + empty endpoint, so existing devices keep waking via FCM with no migration. The Firestore REST wire round-trip persists and reads both fields.

## /register accepts an endpoint in lieu of fcm_token (closes the #469/#470 gap)
`/register` previously required a non-empty `fcm_token`, which foss devices don't have. It now accepts **EITHER** a non-empty `fcm_token` (→ `push_kind=fcm`) **OR** a non-empty UnifiedPush `push_endpoint` (→ `push_kind=unifiedpush`); at least one is required, neither is rejected with 400. `fcm_token` wins if both are present, preserving google behavior. This applies to both the Firebase and keypair register paths (the resolution runs before identity auth).

## UnifiedPush sender
New `UnifiedPushClient` + `RealUnifiedPushClient`: delivers wake/renew by plain HTTP POST to the device's stored endpoint URL — no OAuth, no service account. It reuses the same `FcmData` payload, so the bytes on the wire are exactly `{"type":"wake"}` / `{"type":"renew"}` and the app's existing `FcmDispatch.resolve` handles both transports uniformly. The client is always wired (it needs no credentials).

## Dual routing
New `push::dispatch_push` routing layer selects FCM or UnifiedPush per the device's `push_kind` (returning `NoTarget` when the matching target field is empty). Wired into:
- `passthrough.rs` wake — behind the unchanged FCM wake throttle/retry gate; only the transport swaps.
- `maintenance.rs` renew nudge — same routing, not high-priority.

## Generalized inbound WS control message
The WS handler now accepts the generalized `{"type":"push_endpoint","kind":"fcm"|"unifiedpush","value":...}` shape **and** still accepts the legacy `{"type":"fcm_token","token":...}` shape (the app half will start sending the new shape later). The parsed update sets `push_kind` + the matching target field.

## Tests
- Firestore: UnifiedPush record round-trip; legacy doc with no push fields defaults to FCM.
- `/register`: push_endpoint without fcm_token succeeds and stores `unifiedpush`; fcm_token still works (existing tests); neither → 400. Plus pure-helper unit tests.
- Routing: `dispatch_push` selects the correct client per `push_kind`; `NoTarget` when target empty.
- UnifiedPush sender POSTs `{"type":"wake"}` / `{"type":"renew"}` to the endpoint (captured by a local HTTP server) and surfaces non-2xx as an error.
- WS parsing for both the legacy `fcm_token` and new `push_endpoint` shapes (incl. bad/empty/unknown cases).
- Maintenance: renew nudge for a `unifiedpush` device routes to the UnifiedPush client.

`cargo fmt --check`, `cargo test`, and `cargo clippy --all-targets --all-features -- -D warnings` all clean.

Part of #460 and #463 (relay half; app/foss + docs/ux-decisions entry follow in a second PR).


monkopedia-reviewer (changes requested):
## Review: SSRF in UnifiedPush endpoint — request_changes

**SSRF vulnerability (blocking):** The relay accepts a device-supplied `push_endpoint` URL at `/register` and over the WS `push_endpoint` control message, then POSTs to it on every wake/renew cycle. There is no validation of the URL before it is stored or used. On a multi-tenant relay, an authenticated-but-malicious device can set the endpoint to an internal or metadata target and cause the relay to issue outbound requests there:

- `http://169.254.169.254/computeMetadata/v1/` — GCP instance metadata
- `http://127.0.0.1:<port>/` — relay-host loopback services
- `http://10.0.0.1/` / `http://192.168.x.x/` — RFC1918 private networks
- `https://metadata.google.internal/` — internal cloud DNS names

The vulnerable paths (no validation, just non-empty check):

**`relay/src/api/register.rs` — `resolve_register_push_target` (lines 41-52):**
```rust
if let Some(endpoint) = push_endpoint.filter(|e| !e.is_empty()) {
    return Ok((PushKind::UnifiedPush, String::new(), endpoint.to_string()));
    // ↑ stored as-is, no URL validation
}
```

**`relay/src/api/ws.rs` — `parse_push_update` (lines 174-179):**
```rust
"push_endpoint" => {
    if msg.value.is_empty() {
        return None;
    }
    PushKind::from_wire(&msg.kind).map(|kind| (kind, msg.value.clone()))
    // ↑ value stored as-is for UnifiedPush kind, no URL validation
}
```

**`relay/src/unifiedpush.rs` — `RealUnifiedPushClient::send_data_message` (lines ~1148-1167):**
```rust
let resp = self.http.post(endpoint).json(data).send().await ...
// ↑ POSTs to whatever URL was stored
```

### Required fix

Add a `validate_push_endpoint(endpoint: &str) -> Result<(), &'static str>` function (suggest placing it in `unifiedpush.rs` alongside the client abstraction) that enforces at minimum:

1. **Scheme must be `https`** — reject `http://`, `file://`, etc.
2. **Reject loopback:** `127.0.0.0/8`, `::1`, hostname `localhost`
3. **Reject link-local / cloud-metadata:** `169.254.0.0/16` (covers `169.254.169.254`), `fe80::/10`
4. **Reject RFC1918 private:** `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`
5. **Reject internal DNS names:** `.internal`, `.local`, `.localhost` suffixes (covers `metadata.google.internal`, `metadata.gke.internal`, etc.)
6. **Reject IPv4-mapped IPv6** (e.g. `::ffff:192.168.1.1`) — check `Ipv6Addr::to_ipv4_mapped()` and apply the same v4 rules.

Use `reqwest::Url::parse` (already in deps) + `std::net::IpAddr` — no new crate needed.

Call `validate_push_endpoint` in:
- `resolve_register_push_target` before accepting the UnifiedPush branch
- `parse_push_update` for the `"push_endpoint"` + `UnifiedPush` case before returning the update

Add unit tests for the validation function covering: `http://` rejected, `169.254.169.254` rejected, `127.0.0.1` rejected, `10.x` / `192.168.x` / `172.16.x` rejected, `localhost` rejected, `*.internal` rejected, `::1` rejected, `::ffff:10.0.0.1` rejected; and `https://ntfy.sh/x` accepted.

### Everything else: looks good

- **FCM backward-compat:** `#[serde(default)]` on `push_kind` + `unwrap_or_default()` in the Firestore reader correctly defaults legacy docs to `Fcm`. The `legacy_doc_without_push_fields_defaults_to_fcm` test covers this.
- **`/register` routing:** `fcm_token` wins if both supplied; neither → 400; correct.
- **`dispatch_push`:** clean discriminator, returns `NoTarget` on empty target for its kind.
- **WS legacy shape:** `{"type":"fcm_token","token":"..."}` still handled correctly.
- **Test coverage:** routing, sender, WS parsing, Firestore round-trip, maintenance nudge routing — all real and non-tautological. The `spawn_capture_server` pattern in `unifiedpush.rs` tests is solid.
- **Style:** idiomatic, well-commented, consistent with surrounding code.

Fix the SSRF validation and this is mergeable.

monkopedia-reviewer:
**Coder follow-up:** SSRF fix pushed (466f677d).

`validate_push_endpoint` added to `unifiedpush.rs` — requires `https`, rejects loopback (`127.0.0.0/8`, `::1`, `localhost`), link-local/metadata (`169.254.0.0/16`, `fe80::/10`), RFC1918 private ranges, `.internal`/`.local`/`.localhost` DNS names, and IPv4-mapped IPv6. Called from `resolve_register_push_target` (registration) and `parse_push_update` (WS control message, UnifiedPush case). Ten new unit tests cover all rejection categories and the accept path. `cargo test` + `cargo clippy` + `cargo fmt --check` all clean.

Note: the bracket-stripping (`strip_prefix('[').and_then(strip_suffix(']'))`) is needed because `reqwest::Url::host_str()` includes `[…]` brackets for IPv6 literal addresses, which would otherwise silently defeat `IpAddr::parse`.

monkopedia-reviewer (approved):
## Independent Security Review — SSRF Mitigation Complete

This is an independent re-review of the SSRF fix. The previous `request_changes` review was authored by the same reviewer who then wrote the fix; this review verifies that fix independently.

### SSRF checklist — all four layers confirmed present and correct

**1. Redirects disabled** (`Policy::none()` in `build_client()`)
`RealUnifiedPushClient::build_client()` always passes `redirect::Policy::none()`. This is called from both `new()` (production, no pin) and from `guarded_client()` (with the validated pin). A 3xx to an internal address cannot be followed. Verified by the `redirect_is_not_followed` test.

**2. Resolve-time IP validation on every send**
`send_data_message()` calls `self.guarded_client(endpoint)` unconditionally when `self.enforce_guard` is `true`. Production `new()` always sets `enforce_guard: true`. `guarded_client` resolves via `tokio::net::lookup_host` and rejects if **any** resolved address is blocked — not just the first — closing the multi-A-record edge case. Verified by the `with_resolver` injector tests.

**3. Connection pinning (no TOCTOU window)**
After iterating over all addresses and confirming all are public, `guarded_client` returns `Self::build_client(Some((host, addrs[0])))`. `addrs[0]` is the address just validated. `ClientBuilder::resolve(host, addr)` overrides DNS for that host in the returned client, so no subsequent re-resolve can substitute a different IP between the check and the connect.

**4. `is_blocked_ip` range coverage — no gaps found**
- Loopback: `v4.is_loopback()` (127/8) + `v6.is_loopback()` (::1) ✓
- Link-local / cloud metadata: `v4.is_link_local()` (169.254/16) + `is_ipv6_link_local()` (fe80::/10) ✓
- RFC 1918: `v4.is_private()` (10/8, 172.16/12, 192.168/16) ✓
- IPv4-mapped IPv6: `v6.to_ipv4_mapped()` with full v4 checks ✓
- IPv6 ULA: `is_ipv6_unique_local()` (fc00::/7) ✓
- Unspecified: `v4.is_unspecified()` + `v6.is_unspecified()` ✓

**5. https-only enforced; non-http schemes (`file://`, `ftp://`) rejected at the scheme check.**

**6. Test-only relaxations are `#[cfg(test)]`-gated**
`unguarded()` and `with_resolver()` live in a `#[cfg(test)] impl` block and cannot be called in production. `new()` (the only non-test constructor) always sets `enforce_guard: true` with a real `tokio::net::lookup_host` resolver.

### Edge cases considered and found non-exploitable
- **Trailing-dot hostname** (`metadata.google.internal.`): bypasses the suffix string check but still resolves to 169.254.x.x in GCP → caught by resolve-time IP check. Defense is layered correctly.
- **Octal/hex/decimal IPv4 in URL**: the `url` crate (WHATWG-compliant) normalizes these during parsing → `host_str()` returns canonical decimal dotted form → caught by `IpAddr::parse` + `is_blocked_ip`. Even if not normalized, `lookup_host` returns the canonical `IpAddr` → caught at resolve time.

### Rest of PR
- Additive model (`#[serde(default)]` on `push_kind`/`push_endpoint`) is backward-compatible; legacy FCM documents deserialize correctly as verified by the `legacy_doc_without_push_fields_defaults_to_fcm` test.
- `/register` push-target resolution correctly prioritizes `fcm_token` over `push_endpoint`, with `validate_push_endpoint` called before storing.
- `dispatch_push` correctly dispatches per `push_kind` and returns `NoTarget` on empty target.
- WS `parse_push_update` accepts both legacy and new shapes, calls `validate_push_endpoint` for UnifiedPush kind.
- Tests are substantive (not tautological); Firestore round-trip, routing, WS parsing, and send-path tests all present.

CI: Build ✓, Lint ✓, relay tests ✓, integration tests ✓ on HEAD `f0a19129`.

**Verdict: SSRF mitigation is complete and not bypassable. Approving.**

monkopedia-reviewer (approved):
Independent re-review: SSRF mitigation complete. All four layers verified: redirects disabled (Policy::none always in build_client), resolve-time IP check on every send (enforce_guard:true in production new()), connection pinning to validated addr (build_client with resolve override), is_blocked_ip covers loopback/link-local/RFC1918/IPv4-mapped-IPv6/ULA/unspecified. Test-only unguarded()/with_resolver() properly cfg(test)-gated. No exploitable gaps found. CI green.

PR #473, opened 2026-06-13
FOSS #463: UnifiedPush wake path (app/UX half, deferred activation)

### LINKED ISSUE #463: FOSS #3: UnifiedPush wake path (foss app + relay), dual-routed with FCM

monkopedia-coder:
App/UX half of #463 — the foss-flavor UnifiedPush on-demand wake path. The relay half (#472) is already merged; this PR is app-only (no relay changes).

## Resolved decision — option A (deferred activation)
A foss device has no push endpoint until a UnifiedPush distributor is chosen, and the merged relay's `/register` **requires** a push target. So foss registration is **deferred** until a distributor is selected (Jason, in-session 2026-06-13):

- **Onboarding completes without registering.** foss `OnboardingViewModel.startOnboarding` marks an onboarding-complete flag and lands on Home instead of calling `/register` (which would be rejected with no endpoint). The google flow is unchanged.
- **Home gate relaxed for foss.** `MainActivity` + `OnboardingViewModel.init` treat a foss device with the onboarding-complete flag (but no subdomain) as a valid *degraded* onboarded state, rather than bouncing back to onboarding forever. google still gates on subdomain only.
- **Skip → degraded Home + banner.** "Skip for now" completes onboarding to a degraded Home showing "On-demand wake is off — set up a delivery app".
- **Pick a distributor → activate.** Choosing a delivery app (from the picker, banner, or Settings) calls `UnifiedPush.saveDistributor` + `registerApp`; the endpoint arrives async at `UnifiedPushReceiver.onNewEndpoint`, which reports it to the relay — registering the device on the first endpoint (`push_endpoint` at `/register`) and flipping the banner to active. Subsequent endpoints refresh over the tunnel WS (`{"type":"push_endpoint",...}`).

## What's built
- **Flavor seam** `BackgroundDelivery` (main) — google binds `NoOpBackgroundDelivery` (`isSupported=false`, no surfaces); foss binds `UnifiedPushBackgroundDelivery` (wraps `org.unifiedpush.android:connector`, kept out of shared code).
- **foss `UnifiedPushReceiver`** routes wake/renew through a shared `WakeDispatcher` extracted from `FcmReceiver` (both flavors share the integrations-enabled gate + service-start/renew-enqueue), plus `onNewEndpoint` / `onRegistrationFailed` / `onUnregistered`.
- **"Background delivery" picker** (after Welcome; foss-only nav insertion): enumerates installed distributors, **ntfy suggested only when zero installed**, "install another app" deep-link, **re-scan on `ON_RESUME`**. Built on the real design system from the approved `FossDeliveryMockTest` mocks.
- **Settings row** (foss) reopening the picker with the active distributor marked, and a **degraded Home banner** mirroring the cert/battery/notification banner pattern.
- **core/tunnel**: `RegisterRequest.push_endpoint`, `TunnelClient.sendPushEndpoint`, threaded through `OnboardingFlow`/`register`/`registerWithKeypair`. `CertificateStore` onboarding-complete flag.
- **docs/ux-decisions.md**: the approved "Background delivery" entry with the corrected deferred-activation wording (replaces the 2026-06-11 draft's "serves foregrounded meanwhile" promise).

## Verification
- `:app:assembleFossDebug` `:app:assembleGoogleDebug` `:app:testGoogleDebugUnitTest` `:app:testFossDebugUnitTest` — green.
- `ktlintCheck` `detekt` — clean (detekt scans the foss source set).
- New foss screenshot goldens recorded + `verifyRoborazzi{Google,Foss}Debug` — green.
- `:app:integrationTest` (google) + `:app:fossIntegrationTest` — green; the foss keypair round-trip now registers via empty `fcm_token` + `push_endpoint` (the real foss path) against the merged relay.
- `:work` `:core:tunnel` `:core:bridge` tests — green (FcmReceiver refactor + new TunnelClient method).

## For Jason to scrutinise
- First-run flow change (foss): Welcome → **Background delivery** → NotificationPreferences → Home, with registration deferred off "Continue". google onboarding untouched.
- An integration added on a *skipped* (unregistered) foss device: `IntegrationSetupViewModel.awaitRegistrationIfNeeded` waits for registration, so it would block until a delivery app is set up. Left as-is (the banner nudges delivery setup first) — flagging in case you'd prefer a guard.
- Endpoint rotation while the tunnel is disconnected re-reports only on the next `onNewEndpoint`/reconnect path; the relay retains the registered endpoint meanwhile (documented as acceptable).

Fixes #463.


monkopedia-reviewer (commented):
Reviewer-bot (monkopedia-reviewer) — tier 3 hand-off to @Monkopedia.

## What was vetted to merge standard

**CI:** Build & Test ✓, Lint ✓, test (3 integration suites) ✓ (33 min) on 029c9fa6.

### 1. Google flavor truly unchanged
- `NoOpBackgroundDelivery` (`isSupported=false`) gates every new surface. In `OnboardingViewModel.init` the `backgroundDelivery.isSupported && ...` branch is short-circuit-false on google; `startOnboarding` returns at the `isSupported` guard so google still registers at the "Continue" tap; in `MainActivity` the new `onboarded` expression collapses to `hasSubdomain` because `isSupported=false`. Verified against the pre-PR gate. `SettingsViewModel` and `MainDashboardViewModel` both gate behind `isSupported`/`NotApplicable`, so no picker row, no degraded banner appears on google.

### 2. Foss degraded state is stable — no bounce loop
A foss user who taps "Skip for now": `completeWithoutRegistration` calls `markOnboardingComplete()` (file-marker) and sets `OnboardingState.Onboarded`. On next launch `OnboardingViewModel.init` hits `backgroundDelivery.isSupported && certificateStore.isOnboardingComplete()` → true → `Onboarded` (no subdomain needed). `MainActivity` gate: `hasSubdomain || (true && true)` = true → HOME. No bounce. Two dedicated unit tests lock this (`foss returning user with onboarding-complete flag lands Onboarded`; `foss onboarding completes without registering and marks onboarding complete`).

### 3. WakeDispatcher extraction is behavior-preserving
`FcmReceiver` now delegates to `WakeDispatcher(this, providerRegistry).dispatch(message.data)`. `WakeDispatcher` contains the identical `dispatchIfIntegrationsEnabled` + `awaitReadyBlocking` + service-start/renew-enqueue logic, with the same 2 s registry-ready timeout. `FcmReceiverTest` continues to exercise the full FCM path through Robolectric and its fake-registry timeout override still works (the override sits in the fake, not the constant). The `REGISTRY_READY_TIMEOUT_MS` constant moved from `internal` on `FcmReceiver` to `public` on `WakeDispatcher` — correct location now that both flavors share it.

### 4. Deferred registration wiring
`onNewEndpoint` → `delivery.onEndpoint(endpoint)` → `registerMutex.withLock` → if no subdomain, `registerWithEndpoint` (calls `onboardingFlow.execute("", pushEndpoint=endpoint)`) → on Success: `registrationStatus.markComplete()` + `Active`. If subdomain already present (previously registered): `refreshEndpoint` → `sendPushEndpoint` over WS if tunnel up, deferred if down. Mutex prevents double-register race. `onRegistrationFailed`/`onUnregistered` reset to `NeedsSetup` only when no subdomain (a registered device stays Active). `UnifiedPushReceiver` injects the concrete `UnifiedPushBackgroundDelivery` type (Koin registers it as both the concrete type and the `BackgroundDelivery` seam — the `bind BackgroundDelivery::class` call does both).

### 5. Picker matches the approved design
`DistributorOptionsBuilder.build`: ntfy suggested only when `installed.isEmpty()`; otherwise lists exactly the installed apps + "Install another app" with no ntfy injection. Four unit tests cover the empty/some-installed/active-marked/non-ntfy-only cases. ON_RESUME rescan is a `DisposableEffect` on `Lifecycle.Event.ON_RESUME` in the destination. Roborazzi goldens recorded for none/installed/active-settings picker, degraded banner, Settings row (both dark/light).

### 6. docs/ux-decisions.md entry
Present and thorough — covers picker-not-wizard, ntfy-only-when-empty, deferred-activation wording (corrects the 2026-06-11 draft's "serves foregrounded meanwhile" language), ON_RESUME rescan, Settings re-entry, and alternatives considered. Satisfies `.claude/rules/ux-changes.md`.

### 7. Coroutine rules
`UnifiedPushBackgroundDelivery` uses the injected `appScope` (no `GlobalScope`). All mutable state updates go through a `Mutex`-protected coroutine. No `runBlocking` in production. `sendPushEndpoint` in `TunnelClientImpl` uses the same string-interpolated JSON pattern as the existing `sendFcmToken` (kind is a constant, value is a URL — no JSON-escape risk). Minor cosmetic note (not blocking): the `_activation` `StateFlow` initialises to `NeedsSetup` before the `init` coroutine can set it to `Active` for already-registered returning foss users; the `onStart { emit(NotApplicable) }` in `MainDashboardViewModel` helps but doesn't fully eliminate a sub-millisecond banner flash on startup. Acceptable given how fast the init coroutine runs.

---

## Three items for @Monkopedia's judgment

These are not correctness bugs — the PR is otherwise tier-1-clean. They are UX calls the author explicitly flagged.

**1. First-run flow reorder (foss):** Welcome → **Background delivery** → NotificationPreferences → Home, with registration deferred off "Continue." The `docs/ux-decisions.md` entry is present and the design is from the approved 2026-06-11 mocks + 2026-06-13 deferred-activation correction. This is the main first-run UX change needing your sign-off before merge.

**2. Skip-then-add-integration:** A foss user who tapped "Skip for now" (unregistered, no subdomain, degraded Home) and then goes to add an integration will hit `IntegrationSetupViewModel.awaitRegistrationIfNeeded`, which blocks until registration completes. Since there is no subdomain yet, it will block indefinitely until they set up a delivery app. The banner nudges delivery setup, but no guard or redirect to the picker is in place. The author asks: should this case get an explicit guard (e.g. detect `isSupported && !registered` and route to the picker instead of waiting)?

**3. Endpoint rotation while disconnected:** If the UnifiedPush distributor rotates the endpoint while the tunnel is not connected, `refreshEndpoint` defers (logs a warning, returns). The relay retains the previously-registered endpoint until the next reconnect, at which point the new endpoint re-reports. The author documents this as acceptable; the relay still knows the old endpoint so wake delivery continues.

---

All correctness checks pass. CI is green. Flagging for your review and merge per tier-3 policy.

Monkopedia (approved):

PR #475, opened 2026-06-14
FOSS #464: ACRA crash reporting (foss) + relay /crash → GitHub issues

### LINKED ISSUE #464: FOSS #4: ACRA crash reporting (foss) + relay /crash receiver -> GitHub issues
Part of #460. Depends on #461 (flavor foundation — done) and is sequenced **after #463** (the foss app/UnifiedPush work) to avoid foss-flavor/`DistributionModule` conflicts — **blocked on #463**.

## Goal

Crash reporting for the `foss` flavor without Crashlytics, using [ACRA](https://github.com/ACRA/acra). Reports land where our Crashlytics reports already do: **GitHub issues**.

## Why not "dump into Firebase from the server"

Crashlytics has **no public server-side ingestion API** — you cannot POST a crash into the Crashlytics dashboard from a backend; ingestion is SDK-only. So forwarding foss crashes into the existing Crashlytics console isn't possible. Instead we converge both pipelines on GitHub issues: we already turn Crashlytics crashes into GH issues (the Claude-for-Chrome Shortcut), so the relay does the same for ACRA reports server-side.

[Acrarium](https://github.com/F43nd1r/Acrarium) (the official ACRA backend) is **rejected**: it's Spring Boot + MySQL, far too heavy for the $5/mo memory-constrained relay VPS.

## Work

**App (`foss` flavor):**
- Add ACRA; configure its `HttpSender` to POST JSON crash reports to a relay endpoint (`POST /crash`).
- Bind `CrashReporter` (`api/.../CrashReporter.kt`) to an ACRA-backed impl in the `foss` source set; `google` keeps `FirebaseCrashReporter`. The interface already exists, so this is a flavor binding swap plus an ACRA config object.
- Respect the existing debug/release collection toggle (`RouseApplication.kt:110`).

**Relay:**
- Add a lightweight `POST /crash` endpoint that accepts ACRA's report JSON. Keep it cheap (memory-constrained box): parse, **sanitize** (strip anything resembling private user data — mirror the sanitization the Crashlytics→issue Shortcut uses), **dedup** by stack-trace fingerprint, and **spike-cap** (don't open N issues for one crash loop). File/append a GitHub issue with the `crash` label.
- Unit-test sanitization, dedup fingerprinting, and spike-cap.

## Acceptance

- A `foss` build's uncaught crash produces a sanitized, deduped GitHub issue via the relay, with no Firebase/Crashlytics dependency in the APK.
- `google` crash reporting unchanged.
- `cargo test`/`clippy` clean; app lint clean.

## Open choice (cheap to defer)

v1 sink = GitHub issues (reuses existing automation conventions). If volume ever warrants a queryable store, a tiny SQLite table on the relay is the memory-cheap upgrade — not Acrarium.

monkopedia-coder:
Part of #460. Fixes #464.

The last implementation ticket of the F-Droid epic: Crashlytics-free crash reporting for the `foss` flavor, converging on GitHub issues (Crashlytics has no server-side ingestion API).

## App (foss flavor only)
- Adds **ACRA** (`ch.acra:acra-core` + `acra-http` 5.13.1) as `fossImplementation`. The `google` flavor never links it.
- Replaces the foss NoOp `CrashReporter` with **`AcraCrashReporter`** (forwards `logCaughtException`/`log`/`setCollectionEnabled` to `ACRA.errorReporter`). `google` keeps `FirebaseCrashReporter`; both sit behind the same Koin seam.
- ACRA needs `attachBaseContext` init (it forks a sender process), so `RouseApplication.attachBaseContext` now calls a flavor-specific **`CrashReporterInitializer`**: foss runs `ACRA.init(...)` with an `HttpSender` POSTing **JSON** to the relay's `POST /crash`; google is a no-op (Crashlytics self-initializes via its plugin ContentProvider).
- Respects the existing debug/release collection toggle (`RouseApplication.kt` `setCollectionEnabled(!isDebugBuild)`): ACRA collection is disabled in debug builds.

## Relay (`POST /crash` pipeline)
New `crash` module + endpoint. For each report: **sanitize** → **dedup** → **spike-cap** → **global cap** → file/append a GitHub issue labeled `crash`.
- **Sanitize**: field allowlist (only `APP_VERSION_*`, `ANDROID_VERSION`, `PHONE_MODEL`, `BRAND`, `PACKAGE_NAME` + the stack trace survive — `INSTALLATION_ID`/`DEVICE_ID`/`USER_EMAIL`/`LOGCAT`/etc. are dropped), plus regex redaction of emails / IPs / on-device paths / long opaque tokens in the stack trace.
- **Dedup**: stack-trace fingerprint (exception class + top frames, line numbers stripped, message dropped → stable across builds). Same fingerprint appends a comment instead of opening a new issue.
- **Spike-cap**: caps reports-per-fingerprint per window so one crash loop can't append forever.

## Security / abuse surface
`/crash` takes unauthenticated external POSTs and (indirectly) files GitHub issues — an abuse vector (cf. the #472 SSRF lesson). **Chosen protection: aggressive rate-limiting + capping, not device auth.** Device mTLS auth is intentionally NOT required — ACRA's `HttpSender` can't present the device client cert, and crashes must be reportable before a device is even provisioned. Instead, three in-memory layers:
1. **Per-IP token-bucket rate limit** (~10/min, burst 20) → `429` on flooding. (Plumbs the TCP peer IP into the relay-API router as a `ClientIp` extension.)
2. **Per-fingerprint spike-cap** → one crash loop ⇒ at most N touches/window.
3. **Global new-issue cap** (the real backstop) → bounds *new* issues/window even under IP/fingerprint rotation.

Responses are coarse (`202` for anything ingested, `400` no-stack-trace, `429` rate-limited, `413` too-large) so a caller can't probe dedup/cap state. Maps are swept by the existing rate-limit sweep loop (bounded memory on the $5/mo VPS).

## Ops dependency — GitHub token (Jason provisions)
Filing needs a GitHub token with `issues:write` on the repo:
```toml
[github]
crash_repo = "Monkopedia/rouse-context"
token_env  = "GITHUB_CRASH_TOKEN"   # name only; token set in /etc/rouse-relay/env
```
**If `crash_repo` is empty or the token env var is unset, the endpoint still sanitizes + dedups + logs and returns 202 — it just skips filing (graceful no-op, never crashes).** Documented in `relay/README.md` + `relay.example.toml`.

## Tests
- Relay unit (`src/crash.rs`): sanitization, fingerprint stability/divergence, allowlist drops identifying fields, dedup create→append, spike-cap, window reset, global cap, sweep eviction, per-IP rejection, filing orchestration via a fake filer.
- Relay endpoint (`tests/crash_test.rs`): invalid JSON → 400, missing stack → 400, no-sink → 202, filer-called → 202, per-IP burst → 429.
- App: `testFoss` smoke test for `AcraCrashReporter` (degrades gracefully pre-init).

## Verification
- Relay: `cargo fmt --check` clean, `cargo test` (280+ tests incl. new crash tests) green, `cargo clippy --all-targets --all-features -- -D warnings` clean.
- App: `:app:assembleFossDebug` + `:app:assembleGoogleDebug` + `:app:testGoogleDebugUnitTest` + `:app:testFossDebugUnitTest` + `ktlintCheck` + `detekt` all green.
- Confirmed `firebase-crashlytics` is absent from the foss runtime classpath (present in google). The foss CrashReporter binding resolves to the ACRA impl.

## Note (out of scope for #464)
The foss runtime classpath still pulls `firebase-auth`/`firebase-messaging` transitively via the `:work` module (`FirebaseRenewalAuthProvider`) — pre-existing, not touched here. That's the broader epic's module-decoupling concern (the #462 keypair-auth/`:work` cleanup), not crash reporting. **Crashlytty is fully removed from foss; this PR's scope is met.**

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -253,6 +253,15 @@ dependencies {
     // flavor-agnostic BackgroundDelivery seam.
     "fossImplementation"(libs.unifiedpush.connector)
 
+    // ACRA — scoped to the `foss` flavor only (issue #464). Crashlytics is not
+    // FOSS, so the foss flavor reports crashes via ACRA's HttpSender, which
+    // POSTs JSON crash reports to the relay's `POST /crash` endpoint. The relay
+    // sanitizes/dedups them into GitHub issues (mirroring the Crashlytics→issue
+    // convention). The google flavor never links ACRA — it keeps Crashlytics
+    // behind the same flavor-agnostic CrashReporter seam.
+    "fossImplementation"(libs.acra.core)
+    "fossImplementation"(libs.acra.http)
+
     // Health Connect (for HealthConnectIntegration availability check)
     implementation(libs.health.connect)
 
--- a/app/src/foss/java/com/rousecontext/app/di/DistributionModule.kt
+++ b/app/src/foss/java/com/rousecontext/app/di/DistributionModule.kt
@@ -9,6 +9,7 @@ import com.rousecontext.app.auth.NoOpFcmTokenProvider
 import com.rousecontext.app.delivery.BackgroundDelivery
 import com.rousecontext.app.delivery.UnifiedPushBackgroundDelivery
 import com.rousecontext.app.state.DeviceRegistrationStatus
+import com.rousecontext.app.support.AcraCrashReporter
 import com.rousecontext.tunnel.CertificateStore
 import com.rousecontext.tunnel.OnboardingFlow
 import com.rousecontext.tunnel.TunnelClient
@@ -26,11 +27,14 @@ import org.koin.dsl.module
  * registration and provisioning go through [KeypairDeviceCredentialProvider]
  * (signed registration proof, no Firebase), and expired-cert renewal goes
  * through [KeypairRenewalAuthProvider] (keypair-signed timestamp proof). Crash
- * reporting (#464) and push (#463) remain stubbed until their tickets land.
+ * reporting is ACRA-backed (issue #464); push (#463) remains stubbed until its
+ * ticket lands.
  */
 val distributionModule = module {
-    // Crash reporting — no-op until a FOSS crash backend lands (#464).
-    single { CrashReporter.NoOp }
+    // Crash reporting — ACRA, the FOSS Crashlytics replacement (#464). ACRA is
+    // initialized in CrashReporterInitializer (attachBaseContext); this binding
+    // forwards caught-exception logging / collection toggling to it.
+    single<CrashReporter> { AcraCrashReporter() }
 
     // Push-token retrieval — foss has no FCM token; its push target is the
     // UnifiedPush endpoint (reported via BackgroundDelivery), so the FCM-token
--- a/app/src/foss/java/com/rousecontext/app/support/AcraCrashReporter.kt
+++ b/app/src/foss/java/com/rousecontext/app/support/AcraCrashReporter.kt
@@ -1,0 +1,40 @@
+package com.rousecontext.app.support
+
+import com.rousecontext.api.CrashReporter
+import java.util.concurrent.atomic.AtomicLong
+import org.acra.ACRA
+
+/**
+ * [CrashReporter] backed by ACRA (the FOSS Crashlytics replacement), used by
+ * the `foss` flavor. The parallel `google` flavor binds
+ * [com.rousecontext.app.support.FirebaseCrashReporter]; both sit behind the
+ * same Koin seam. Issue #464.
+ *
+ * ACRA is initialized in [CrashReporterInitializer] from
+ * `Application.attachBaseContext`. Before that runs (or if init is skipped,
+ * e.g. in the ACRA sender process), [ACRA.errorReporter] returns a safe no-op
+ * stub, so every method here degrades gracefully rather than throwing.
+ */
+class AcraCrashReporter : CrashReporter {
+
+    /** Monotonic counter so breadcrumb custom-data keys never collide. */
+    private val breadcrumbSeq = AtomicLong(0)
+
+    override fun logCaughtException(throwable: Throwable) {
+        // Non-fatal: report without killing the process, mirroring
+        // Crashlytics.recordException.
+        ACRA.errorReporter.handleSilentException(throwable)
+    }
+
+    override fun log(message: String) {
+        // Attach as custom data so it rides along with the next crash report,
+        // the closest ACRA analogue to a Crashlytics breadcrumb. Keys are
+        // sequenced so successive breadcrumbs don't overwrite each other.
+        val key = "breadcrumb_${breadcrumbSeq.incrementAndGet()}"
+        ACRA.errorReporter.putCustomData(key, message)
+    }
+
+    override fun setCollectionEnabled(enabled: Boolean) {
+        ACRA.errorReporter.setEnabled(enabled)
+    }
+}
--- a/app/src/foss/java/com/rousecontext/app/support/CrashReporterInitializer.kt
+++ b/app/src/foss/java/com/rousecontext/app/support/CrashReporterInitializer.kt
@@ -1,0 +1,69 @@
+package com.rousecontext.app.support
+
+import android.app.Application
+import com.rousecontext.app.BuildConfig
+import org.acra.ACRA
+import org.acra.config.CoreConfigurationBuilder
+import org.acra.config.HttpSenderConfigurationBuilder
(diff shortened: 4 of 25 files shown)


monkopedia-reviewer (changes requested):
**Security review — one blocking finding, otherwise LGTM.**

## What was checked

- CI: Build & Test ✓, Lint ✓, relay tests ✓, integration tests ✓ (all green).
- Dormant-by-default: confirmed. `build_filer_from_config` returns `None` when `crash_repo` is empty OR the token env var is unset/empty. `CrashService::disabled()` is the default in `main.rs` until the operator provisions both. Merging this does not activate an open issue-filer.
- Rate limits & caps: per-IP token-bucket (burst 20, ~10/min), per-fingerprint spike-cap (5/hr), global new-issue cap (20/hr) all confirmed present and tested. All three in-memory maps are swept via `app_state.crash.sweep_expired()` in the existing sweep loop — bounded memory.
- Stack trace injection: in a fenced code block — safe.
- Fingerprint dedup logic: stable across message/line-number variance, correctly resets while retaining issue number, global cap is a genuine backstop against fingerprint rotation.
- Issue-closing keywords in an issue *body* (not a commit/PR): GitHub does not auto-close other issues from keywords in issue bodies — not a real risk here.

## Blocking: `@mention` injection via allowlisted field values and issue title

`sanitize_text` redacts emails, IPv4 addresses, on-device paths, and long tokens, but it does **not** strip or escape `@` characters. The allowlisted field values (`APP_VERSION_NAME`, `APP_VERSION_CODE`, `ANDROID_VERSION`, `PHONE_MODEL`, `BRAND`, `PACKAGE_NAME`) are placed **outside** the fenced code block as markdown list items:

```
- **App version:** <value from request>
```

An unauthenticated external attacker can POST arbitrary field values. If they set `APP_VERSION_NAME` to `@someuser`, the filed GitHub issue body will contain `- **App version:** @someuser`. Since `Monkopedia/rouse-context` is a **public** repo, GitHub sends a notification email to `@someuser` for every such issue.

The issue title has the same exposure. `build_title` extracts the first line of the (sanitized) stack trace, splits on `:` to get the "class", then takes the component after the last `.`. If the attacker sends a stack trace whose first line contains no `:` and no `.` (e.g. `@evil`), `short = "@evil"` and the title becomes `[crash] @evil (<hash>)`. GitHub @mention notifications fire from issue titles too on a public repo.

With the global cap at 20/hr, an attacker can trigger 20 GitHub notification emails per hour to arbitrary GitHub users. After a restart, the cap resets (in-memory only), so the sustained rate per restart is bounded but an attacker who can trigger restarts gets a re-burst.

**Fix:** extend `sanitize_text` to also replace bare `@` with a non-triggering form — the conventional choice is inserting a zero-width space (`@\u{200B}`) so the display is unchanged but GitHub doesn't interpret it as a mention. Apply to the allowlisted-field path too (which already calls `sanitize_text`, so this would be a one-line addition to that function). Alternatively put all field values inside the code block, but the zero-width-space approach is less disruptive.

The stack trace is already sanitized with `sanitize_text` before being passed to `build_title` and `fingerprint`, so the same fix covers the title path.

## Not blocking (note only)

The pre-existing `firebase-auth`/`firebase-messaging` transitive pull via `:work` in the foss classpath is out of scope for #464 and already tracked as #476. Confirmed: not touching this.

## Everything else

App flavor wiring (ACRA in `fossImplementation` only, google flavor untouched), `attachBaseContext` init path, debug/release gate, `AcraCrashReporter` graceful-pre-init behaviour, Koin seam — all sound. Tests are real (not tautological): sanitize, fingerprint divergence, dedup create→append→spike-cap→window-reset, global cap, sweep eviction, per-IP rejection, filing orchestration, endpoint status codes. `cargo fmt`/clippy clean per CI.

Ready to approve and merge once the `@mention` sanitization is added.

monkopedia-reviewer (approved):
The @mention / #ref injection finding from the prior review is resolved. Commit 0af50bda adds the zero-width-space defang (s.replace('@', "@\u{200B}").replace('#', "#\u{200B}")) at the return of sanitize_text, which is applied to both the stack trace and all allowlisted body fields before they reach the GitHub issue title/body. A dedicated unit test (sanitize_neutralizes_github_mentions_and_refs) confirms both transforms. The fix is scoped to relay/src/crash.rs only — no unrelated changes. CI green on HEAD (Build & Test, Lint, both test jobs). Everything else (dormant-by-default, three rate-limit layers, map sweeping) was vetted in the prior review and is unchanged. Approving.

PR #479, opened 2026-06-15
#467: replace google/foss product flavors with a -Pgoogle build flag

### LINKED ISSUE #467: Harden FOSS flavor: replace conditional google-services plugin application with hand-rolled Firebase config resources
Follow-up from #461/#466 (the `google`/`foss` flavor foundation).

## Context
#466 keeps the `foss` build Firebase-free by **conditionally applying** the `google-services` + `firebase-crashlytics` Gradle plugins based on the requested task name (apply only when a `google`-flavor task is built). This is the standard FOSS-flavor pattern, but it has a footgun:

- **`./gradlew :app:assemble` (both flavors at once) still applies the plugins** → the `foss` APK from a bare `assemble` is NOT actually Firebase-free. CI and F-Droid invoke the specific flavor task (`assembleFossRelease`/`assembleFossDebug`), so the **shipped** foss APK is genuinely clean — but a dev running bare `assemble` locally gets a misleading artifact.

## Cleaner approach to evaluate
Drop the `google-services` plugin entirely and hand-write its output as string resources in `app/src/google/res/values/` — `google_app_id`, `gcm_defaultSenderId`, `project_id`, `google_api_key` (the plugin's only job is generating these from `google-services.json`; the Firebase SDK reads them at runtime). Then `google` has the resources, `foss` doesn't, and there's no conditional or task-name sniffing — and the bare-`assemble` footgun disappears.

**Caveat:** this cleanly removes the conditional for `google-services`, but **Crashlytics is stickier** — its plugin generates a build-id resource the SDK needs and does mapping upload, which doesn't hand-roll as cleanly. So Crashlytics may keep a (smaller) conditional or need separate handling.

## Recommendation
Tackle this **alongside #464** (ACRA / Crashlytics rework), since that ticket already revisits the crash-reporting plugin/SDK for the `foss` flavor — doing both halves together solves google-services + Crashlytics in one pass instead of half now. Until then, the current conditional is correct for the shipped artifacts.

Part of #460.

monkopedia-coder:
## What & why

`Closes #467.`

Replaces the `google`/`foss` **product flavors** with a **`-Pgoogle` build flag** (approved design, Jason in-session 2026-06-15).

FOSS and Google are not symmetric peers: FOSS is the universal, secret-free base build (buildable by anyone / F-Droid); Google is a credentialed *superset* requiring our private `google-services.json`. Product flavors forced a symmetric frame — bare `assemble` built both, `google-services.json` was required by default, and task-name sniffing (`isFossOnlyBuild`) was needed to re-derive "is this the credentialed build?". A flag encodes the real relationship: **bare build = FOSS; `-Pgoogle` = Firebase opt-in.** The safe build becomes the default, and keeping the google-services plugin behind the flag preserves its hard-fail-on-missing-credentials guard.

## The inversion (`app/build.gradle.kts`)

- `val googleBuild = project.hasProperty("google")`.
- Removed `flavorDimensions` / `productFlavors`. Distribution sources selected by flag: `sourceSets["main"]/["test"].java.srcDir(if (googleBuild) "src/google/java"/"src/testGoogle/java" else "src/foss/java"/"src/testFoss/java")`. The `src/{google,foss}` and `src/test{Google,Foss}` directories are unchanged.
- **Manifest overlay** by flag via AGP 8.13.2 Variant Sources API — `androidComponents.onVariants { variant.sources.manifests.addStaticManifestFile(...) }` — feeding `src/{google,foss}/AndroidManifest.xml` into the standard merger as the highest-priority source. No module split, no shared-body duplication.
- Firebase deps + the `google-services`/`firebase-crashlytics` plugins + the Crashlytics `buildTypes` config all gated on `googleBuild` (replacing the task-name sniff). `google-services.json` required **only** under `-Pgoogle`, still hard-failing loudly if absent.
- FOSS deps (`unifiedpush.connector`, `acra.core`, `acra.http`) on the `else` branch.
- Collapsed `fossIntegrationTest` into a single flag-selected `integrationTest`; fixed the release-signing task set (`assembleRelease`/`packageRelease`/`bundleRelease`) and the root `build.gradle.kts` Kover disabled-task list, both of which named the now-gone per-flavor tasks.

## CI two-pass migration

Per-flavor tasks (`testGoogleDebugUnitTest`, `assembleFossDebug`, `fossIntegrationTest`, …) no longer exist, so both distributions are covered by running twice per flag:

- **android-ci.yml / ci.yml** — `assembleDebug` (FOSS) + `assembleDebug -Pgoogle`; `:app:testDebugUnitTest` (FOSS) + `-Pgoogle`; ktlint runs both passes (`-Pgoogle` lints the Firebase sources); detekt stays single-pass (its `src` glob already scans both dirs).
- **android-ci.yml integration** — `:core:tunnel:integrationTest :app:integrationTest` (FOSS, incl. the keypair round-trip) then `:app:integrationTest -Pgoogle`; Kover runs `-Pgoogle` to keep `:app` coverage on the shipping Firebase distribution (reused UP-TO-DATE).
- **release.yml** — `assembleRelease`/`bundleRelease -Pgoogle`, with the un-prefixed `app/build/outputs/{apk,bundle}/release/app-release.*` paths.
- Docs (`agent-quickstart`, `self-hosting`, `test-coverage-audit`) updated to the flag form.

## Verification (local, JDK 21)

- **FOSS** `:app:assembleDebug` ✅ — merged manifest registers `UnifiedPushReceiver`, **no** `FcmReceiver`/`MESSAGING_EVENT` component; `debugRuntimeClasspath` has **zero** `firebase-*` artifacts and links `unifiedpush` + `acra`.
- **Google** `:app:assembleDebug -Pgoogle` ✅ — merged manifest registers `FcmReceiver` (`MESSAGING_EVENT`), **no** `UnifiedPushReceiver`; classpath has 62 firebase artifacts and **zero** unifiedpush/acra. (Proves `addStaticManifestFile` works both ways.)
- **Both unit-test passes green:** `:app:testDebugUnitTest` (ran `AcraCrashReporterTest`) and `-Pgoogle` (ran `FirebaseRenewalAuthProviderTest`, `FcmReceiverTest`, `FirebaseCrashReporterTest`).
- **Lint green** both passes: `ktlintCheck` + `:app:ktlintCheck -Pgoogle`; `detekt`.
- **Loud-fail guard confirmed:** with `app/google-services.json` temporarily moved, `:app:processDebugGoogleServices -Pgoogle` fails with `File google-services.json is missing.`; the bare FOSS build never invokes that plugin. (File restored.)
- `:app:integrationTest` task wiring resolves under both flags (dry-run).

## Notes / unsure

- `google-services.json` is checked in (`app/google-services.json`; API keys aren't secret), so `-Pgoogle` builds succeed in CI and locally — the loud-fail path is only hit when it's genuinely absent.
- Full relay-backed `integrationTest` suites were not run locally (need the `--features test-mode` relay binary); CI runs them both passes. Task graph validated via dry-run.
- Kover now tracks `:app` coverage under `-Pgoogle` to match the prior google-flavor-only coverage intent; FOSS-only seams (ACRA/keypair) are exercised by tests but not aggregated into the report.

### DIFF
--- a/.github/workflows/android-ci.yml
+++ b/.github/workflows/android-ci.yml
@@ -53,7 +53,13 @@ jobs:
             -alias debug \
             -dname "CN=Debug,O=Debug,L=Debug,S=Debug,C=US"
       - name: Lint
-        run: ./gradlew ktlintCheck
+        # Distribution is a build flag, not a flavor (#467), so a single
+        # ktlintCheck only sees the selected (FOSS, default) `src` dir. Run a
+        # second pass with `-Pgoogle` so the Firebase distribution sources are
+        # linted too. (`:app` is the only module whose sources vary by flag.)
+        run: |
+          ./gradlew ktlintCheck
+          ./gradlew :app:ktlintCheck -Pgoogle
       - name: Check for zombie screen files
         run: bash scripts/check-zombie-screens.sh
       - name: Check for manual-template JSON in production code
@@ -62,14 +68,18 @@ jobs:
         # when interpolated values contain quotes, backslashes, or control
         # characters. Production code must use @Serializable + encodeToString.
         run: bash scripts/check-no-manual-json.sh
-      - name: Build (google flavor)
-        run: ./gradlew :app:assembleGoogleDebug
-      - name: Build (foss flavor, Firebase-free)
-        # Separate invocation so the foss-only build skips the Firebase Gradle
-        # plugins and proves it assembles without google-services. See #461.
-        run: ./gradlew :app:assembleFossDebug
+      - name: Build (Google distribution)
+        run: ./gradlew :app:assembleDebug -Pgoogle
+      - name: Build (FOSS distribution, Firebase-free)
+        # Bare build (no `-Pgoogle`) is the FOSS distribution: it skips the
+        # Firebase Gradle plugins and proves it assembles without
+        # google-services. See #467.
+        run: ./gradlew :app:assembleDebug
       - name: Unit tests
         timeout-minutes: 20
+        # Distribution is a build flag (#467), so `:app` unit tests run twice:
+        # bare = FOSS (ACRA / keypair seams), `-Pgoogle` = Firebase-backed. The
+        # library modules are distribution-agnostic and run once.
         run: |
           ./gradlew \
             :core:tunnel:jvmTest \
@@ -79,7 +89,8 @@ jobs:
             :notifications:testDebugUnitTest \
             :work:testDebugUnitTest \
             :integrations:testDebugUnitTest \
-            :app:testGoogleDebugUnitTest
+            :app:testDebugUnitTest
+          ./gradlew :app:testDebugUnitTest -Pgoogle
       - name: Integration tests (relay)
         if: success() || failure()
         # Raised from 10 → 15 as the integration-scenario coverage grew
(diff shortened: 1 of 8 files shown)


monkopedia-reviewer (approved):
Reviewed to tier-1 standard. All four load-bearing points verified:

**1. `addStaticManifestFile` manifest-overlay wiring** — `androidComponents { onVariants(selector().all()) { variant -> variant.sources.manifests.addStaticManifestFile(if (googleBuild) "src/google/AndroidManifest.xml" else "src/foss/AndroidManifest.xml") } }`. `googleBuild` is resolved at configuration time via `project.hasProperty("google")`, so the correct overlay is always fed into the merger. PR description confirms local verification: FOSS merged manifest registers `UnifiedPushReceiver` (no `FcmReceiver`); Google registers `FcmReceiver` (no `UnifiedPushReceiver`). ✅

**2. CI two-pass matrix coverage** — both workflows (android-ci.yml, ci.yml) run two explicit passes for build, unit tests, ktlint, and integration tests. Library modules (distribution-agnostic) run once; only `:app` runs twice. detekt stays single-pass and already globs both source dirs. No coverage silently dropped. ✅

**3. Firebase firewall** — plugins (`google-services`, `firebase-crashlytics`) gated behind `if (googleBuild)`; dependencies likewise in an `if/else` block; source dir is `src/foss/java` by default. FOSS build never invokes Firebase Gradle processing. ✅

**4. Loud-fail guard preserved** — `google-services` plugin only applied under `-Pgoogle`; its own hard-fail-on-missing-credentials fires when `google-services.json` is absent in that path. FOSS bare build never touches the plugin. ✅

**Additional mechanics verified:**
- `integrationTest` task correctly sources `testDebugUnitTest` classpath, which is already configured with the right distribution at Gradle config time.
- `fossIntegrationTest` properly collapsed: FOSS is now the default `integrationTest` bare, Google is `-Pgoogle`.
- Kover disabled-task list cleaned up (flavor-specific names removed; single `testReleaseUnitTest` covers all modules).
- Release artifact paths updated to un-prefixed AGP output paths (`apk/release/app-release.apk`, `bundle/release/app-release.aab`) — correct for unflavored builds.
- Docs (agent-quickstart, self-hosting, test-coverage-audit) all updated consistently.

CI: all three checks (test 33m, Lint 2m42s, Build & Test 10m22s) completed and passed. Matches the explicitly user-approved design from issue #467.

PR #482, opened 2026-06-16
Redirect unregistered FOSS Add-integration to delivery picker (#474)

### LINKED ISSUE #474: FOSS: adding an integration on a skipped (unregistered) device should redirect to the delivery picker, not block
Follow-up from the #473 review (foss UnifiedPush app/UX).

On the FOSS flavor, a user who taps **"Skip for now"** on Background delivery reaches a degraded, *unregistered* Home (registration is deferred until a distributor is chosen — see the 2026-06-13 ux-decisions entry). If they then try to **add an integration**, `IntegrationSetupViewModel.awaitRegistrationIfNeeded` **blocks/waits** until delivery is set up. The Home banner nudges setup, but the add-integration flow itself just waits with no explanation — reads like a hang.

**Proposed:** when adding an integration on a not-yet-registered foss device, detect the state and **redirect to the Background delivery picker** (or show a clear "set up a delivery app first" prompt) instead of silently blocking.

Context: Jason reviewed #473 and approved it as-is (clean approval); this is the tracked polish, not a blocker. Recommended fix shape is small + app-only. Part of #460.

monkopedia-coder:
Closes #474

## What changed

On a not-yet-registered FOSS device (one that tapped "Skip for now" on Background delivery and sits on a degraded Home), tapping **Add integration** used to silently block in `IntegrationSetupViewModel.awaitRegistrationIfNeeded` waiting for a registration that never starts — it read like a hang.

This implements the approved **Option A + auto-resume**:

1. **Detect + redirect.** At add-integration time, if `BackgroundDelivery.activation == NeedsSetup` (the foss skipped/unregistered state), redirect to the existing #463 Background delivery picker instead of blocking.
2. **Contextual strip.** When the picker is reached this way it shows: *"Set up a delivery app to finish adding integrations."*
3. **Auto-resume.** A new `PendingIntegrationSetup` singleton carries the integration id across the picker round-trip. Once a distributor is chosen, the user drops straight back into that integration's setup flow (the setup screen's existing "Registering" indicator covers the brief deferred-registration wait). Tapping "Skip for now" instead drops the pending id and returns to Add integration.

Gated on `NeedsSetup`, so it is a **no-op on google** (FCM → `NotApplicable`) and on already-registered FOSS devices (`Active`), which proceed straight into setup as before.

The integration-id → setup-route mapping was extracted into a shared `setupRouteForIntegration` helper so the picker and the add-integration screen resolve the same dedicated setup screens (Health Connect, notifications, outreach, usage).

## UX decision log

Per `.claude/rules/ux-changes.md`, this PR adds a `docs/ux-decisions.md` entry (newest on top): Decision = Option A redirect + auto-resume; Approved by Jason in-session 2026-06-15 (reviewed Roborazzi mocks of baseline/A/B/C); records the B/dialog, C/inline-banner, and return-to-home alternatives and the trade-off. Relevant: #474, #463, #460.

## Tests / verification

- New `PendingIntegrationSetupTest`: remember/consume/clear carry semantics + the redirect decision fires only for `NeedsSetup` (not `Active`/`NotApplicable`).
- New `IntegrationSetupRoutesTest`: shared route mapping resolves each specialized + generic setup route.
- `BackStackFlowTest`: two new cases — redirect then resume lands on integration setup (not Home, picker popped); redirect then skip returns to Add integration.
- `./gradlew :app:testDebugUnitTest` green; `./gradlew ktlintCheck detekt` green.
- google variant compiles (`:app:compileDebugKotlin -Pgoogle`). `:app:assembleDebug` fails only at `validateSigningDebug` in this worktree (keystore/env), identically for foss and google — not introduced here.

### DIFF
--- a/app/src/main/java/com/rousecontext/app/di/AppModule.kt
+++ b/app/src/main/java/com/rousecontext/app/di/AppModule.kt
@@ -31,6 +31,7 @@ import com.rousecontext.app.state.DeviceRegistrationStatus
 import com.rousecontext.app.state.IntegrationSettingsStore
 import com.rousecontext.app.state.NotificationPermissionRefresher
 import com.rousecontext.app.state.OAuthHostnameProvider
+import com.rousecontext.app.state.PendingIntegrationSetup
 import com.rousecontext.app.state.PreferencesSnapshotHolder
 import com.rousecontext.app.state.SecurityAlertGate
 import com.rousecontext.app.state.ThemePreference
@@ -175,6 +176,9 @@ val appModule = module {
 
     // --- Notification permission refresher ---
     single { NotificationPermissionRefresher() }
+
+    // --- Pending-integration carry across the Background delivery detour (#474) ---
+    single { PendingIntegrationSetup() }
 
     // --- Bug report URI builder ---
     single { BugReportUriBuilder(androidContext()) }
--- a/app/src/main/java/com/rousecontext/app/state/PendingIntegrationSetup.kt
+++ b/app/src/main/java/com/rousecontext/app/state/PendingIntegrationSetup.kt
@@ -1,0 +1,58 @@
+package com.rousecontext.app.state
+
+import com.rousecontext.app.delivery.DeliveryActivation
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+
+/**
+ * Carries the id of the integration a user was adding across a Background
+ * delivery picker detour (issue #474, Option A).
+ *
+ * On a not-yet-registered foss device (one that tapped "Skip for now" on
+ * Background delivery), tapping "Add integration" used to silently block in
+ * [com.rousecontext.app.ui.viewmodels.IntegrationSetupViewModel]
+ * `.awaitRegistrationIfNeeded` — it read like a hang. Instead the add-integration
+ * destination now [remember]s the chosen integration and redirects to the
+ * picker. Once a distributor is selected, the picker [consume]s the id and
+ * navigates straight into that integration's setup flow (auto-resume), rather
+ * than dropping the user back at Home.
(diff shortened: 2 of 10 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard. All four focus points clear:

**Gate correctness** — `deliveryNeedsSetupBeforeIntegration` is a strict equality check on `DeliveryActivation.NeedsSetup`; `PendingIntegrationSetupTest` explicitly asserts the other two variants (`Active`, `NotApplicable`) return false. Google/FCM and already-registered FOSS are genuine no-ops.

**Pending-integration carry + auto-resume** — `consume()` atomically returns-and-clears the id (main thread, no race). `onChosen` pops the picker inclusive (`popUpTo(BACKGROUND_DELIVERY_BASE) { inclusive = true }`) so back from setup lands on Add integration, not the picker. `onSkip` calls `popBackStack()` when a pending id exists, returning to Add integration. `onChosen`'s else-branch correctly preserves the `settingsMode` fallback. The apparent missing `settingsMode` branch in `onSkip` is safe: the "Skip for now" button is rendered only when `!settingsMode` in `BackgroundDeliveryScreen`, so `onSkip` is never reachable from the settings path.

**Tests** — `PendingIntegrationSetupTest` covers remember/consume/clear semantics and the redirect-fires/does-not-fire decision for all three `DeliveryActivation` values. `IntegrationSetupRoutesTest` locks in all 4 specialized + generic fallback mappings. `BackStackFlowTest` adds both redirect-then-resume (lands on setup, picker gone) and redirect-then-skip (returns to Add integration) cases.

**UX decision log** — entry present in `docs/ux-decisions.md` (2026-06-16, newest on top), accurate description of Option A + auto-resume, names Jason's in-session approval, records alternatives B/C/return-to-home and trade-offs accepted.

CI: Build & Test ✓, Lint ✓, test ✓. No code quality issues.

PR #488, opened 2026-06-16
#483: FOSS Home battery-opt warning + one-tap exemption dialog

### LINKED ISSUE #483: FOSS: background wake silently fails unless app + distributor are battery-opt-exempt (FCM gets this automatically)
## Problem
On the FOSS build, on-demand wake **silently fails from the background** unless the user has granted a battery-optimization exemption to **both** the UnifiedPush distributor (e.g. ntfy) **and** our app, and ideally bumped them to the `active` standby bucket.

Mechanism: a high-priority **FCM** message gets the target app a ~10s temporary power-allowlist (Doze + background-FGS-start exemption), granted by **Google Play Services**, which is a privileged system component holding the permission to do so. A third-party UnifiedPush distributor is an ordinary app and **cannot** grant our app that temporary exemption. So when the distributor broadcasts a wake to our `UnifiedPushReceiver`, our attempt to start the tunnel foreground service from the background is blocked, and the wake is dropped. The `google`/FCM flavor never hits this because GMS hands out the exemption automatically and invisibly.

## Evidence
Found during the 2026-06-16 FCM-vs-ntfy wake-latency benchmark on adolin (`solar-dingo`, Pixel 6 Pro). The ntfy wake path did **not** work at all until both `io.heckel.ntfy` and `com.rousecontext.debug` were deviceidle-whitelisted and set to standby bucket `active`. Once exempt, wake latency was effectively tied with FCM (~0.85s vs ~0.89s median) — i.e. this is a **reliability/setup** gap, not a latency one.

## Why needs-decision (fix has shapes)
- **(a) Request the exemption during foss delivery setup** — when the user picks a distributor on the Background delivery picker (#463), prompt for `REQUEST_IGNORE_BATTERY_OPTIMIZATIONS` (or route to the system battery-opt screen, mirroring the existing #453 Connection-settings "Fix this" pattern), and surface that the distributor *also* needs exempting. Most reliable; adds an onboarding permission surface (UX-approval-gated per `.claude/rules/ux-changes.md`).
- **(b) Detect-and-warn** — detect that wake isn't firing / the app isn't battery-exempt and show a fixable Home banner (like the existing degraded-delivery banner), rather than prompting up front.
- **(c) Document-only** — note it in the picker copy / help. Weakest; leaves wake quietly unreliable.

Any of these is a first-run/permission-timing/error-surface change → in-session approval + a `docs/ux-decisions.md` entry required.

## Notes
- Distinct from #480 (that's the distributor's *stopped* state suppressing the initial REGISTER; this is *battery-optimization* exemption suppressing the FGS start on each wake). Both are foss-only background-delivery reliability gaps; cross-referencing.
- `google`/FCM flavor unaffected.
- Part of #460.

monkopedia-coder:
Closes #483

## What & why
On the FOSS build, on-demand wake silently fails from the background unless the app is on the battery-optimization allowlist. A third-party UnifiedPush distributor can't grant our app the temporary power-allowlist that Play Services hands the FCM build, so the background start of the tunnel foreground service is blocked and the wake is dropped. Nothing surfaced this fixable condition.

### Part 1 — FOSS-only Home battery-optimization warning
- New persistent Home banner ("Battery optimization is on") shown when the app is **not** battery-exempt.
- **Foss-gated at runtime** via `BackgroundDelivery.isSupported` — the same seam the existing "On-demand wake is off" delivery banner uses (`NoOpBackgroundDelivery.isSupported == false` on google/FCM, so it never shows there). No new BuildConfig field needed.
- Exemption is re-read on `ON_RESUME` via a new `batteryExemptFlow(context, triggers)` that mirrors `notificationPermissionFlow`, reusing the dashboard's existing `NotificationPermissionRefresher` ticks, so the banner clears when the user returns from the OS dialog.
- Sits alongside (and can co-exist with) the delivery banner — distinct condition: delivery banner = no distributor chosen; battery banner = distributor set up but the app can't start the FGS from background without the exemption.

### Part 2 — "Fix this" becomes the one-tap OS dialog
- Both the new Home banner action and the **existing #453 Connection-settings "Fix this"** now launch `ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS` (one-tap allow/deny dialog) + `package:` URI, instead of `ACTION_IGNORE_BATTERY_OPTIMIZATION_SETTINGS` (the list).
- Declares `REQUEST_IGNORE_BATTERY_OPTIMIZATIONS` in the **shared** manifest (the Settings "Fix this" exists on both builds). #453 used the list solely to dodge the Google Play restriction on that permission; we don't ship to Play (Google build is GitHub-distributed, FOSS goes to F-Droid, which accepts it — ntfy uses it), so that no longer binds.

### UX decision log
`docs/ux-decisions.md` gets a new 2026-06-16 entry recording both parts and explicitly **superseding the 2026-06-02 #453** "Fix this opens battery-optimization settings" decision.

## Tests
- `BatteryOptimizationTest` updated: asserts the dialog action + `package:` data URI.
- `MainDashboardViewModelTest`: banner shows on foss when not exempt; absent when exempt; absent on google even when not exempt; clears reactively when exemption is granted.

## Verification
- `./gradlew :app:testDebugUnitTest` (foss/default) — green (full suite).
- Affected tests also green under `-Pgoogle`.
- `./gradlew :app:compileDebugKotlin -Pgoogle` — compiles (warning stays foss-gated).
- `./gradlew :app:ktlintCheck detekt` — green.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -26,6 +26,19 @@
     -->
     <uses-permission android:name="android.permission.health.READ_HEALTH_DATA_HISTORY" />
 
+    <!--
+      Lets the "Fix this" battery-optimization action open the one-tap
+      allow/deny dialog (ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS) instead of
+      the settings list the user has to hunt through. On the FOSS build the
+      exemption is required for background wake to fire at all (issue #483); on
+      the Google build it improves wake reliability (#453). Google Play
+      restricts this permission, but we don't ship to Play — the Google build is
+      GitHub-distributed and the FOSS build goes to F-Droid, which accepts it
+      (ntfy uses the same permission). Declared in the shared manifest because
+      the Settings "Fix this" exists on both builds.
+    -->
+    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
+
     <application
         android:name=".RouseApplication"
         android:allowBackup="false"
--- a/app/src/main/java/com/rousecontext/app/di/AppModule.kt
+++ b/app/src/main/java/com/rousecontext/app/di/AppModule.kt
@@ -38,6 +38,7 @@ import com.rousecontext.app.state.ThemePreference
 import com.rousecontext.app.state.notificationPermissionFlow
 import com.rousecontext.app.support.BatteryOptimization
 import com.rousecontext.app.support.BugReportUriBuilder
+import com.rousecontext.app.support.batteryExemptFlow
 import com.rousecontext.app.token.RoomTokenStore
 import com.rousecontext.app.token.TokenDatabase
 import com.rousecontext.app.token.createUnknownClientLabeler
@@ -630,7 +631,12 @@ val appModule = module {
                 triggers = refresher.ticks
             ),
             spuriousWakesFlow = SettingsViewModel.spuriousWakeStatsFlow(get()),
-            deliveryActivation = get<BackgroundDelivery>().activation
+            deliveryActivation = get<BackgroundDelivery>().activation,
+            batteryOptimizationExempt = batteryExemptFlow(
+                context = androidContext(),
+                triggers = refresher.ticks
+            ),
+            deliveryWakeSupported = get<BackgroundDelivery>().isSupported
         )
     }
     viewModel { AddIntegrationViewModel(get(), get(), get()) }
--- a/app/src/main/java/com/rousecontext/app/support/BatteryOptimization.kt
+++ b/app/src/main/java/com/rousecontext/app/support/BatteryOptimization.kt
@@ -2,23 +2,34 @@ package com.rousecontext.app.support
 
 import android.content.Context
 import android.content.Intent
+import android.net.Uri
 import android.os.PowerManager
 import android.provider.Settings
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onStart
 
 /**
  * Battery-optimization status and the user-facing flow to change it.
  *
- * Reliable FCM wake-ups depend on the app being exempt from Doze battery
- * optimization. We surface the current status in Settings (a warning + the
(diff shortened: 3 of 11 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard. All 4 focus points verified:

**1. FOSS gate** — `deliveryWakeSupported = get<BackgroundDelivery>().isSupported` in `AppModule.kt`; ViewModel gates the banner on `deliveryWakeSupported && !batteryExempt`. Default `false` → banner never appears on google/FCM or in tests unless explicitly set.

**2. Live state / ON_RESUME refresh** — `batteryExemptFlow` mirrors `notificationPermissionFlow` exactly: `triggers.onStart { emit(Unit) }.map { BatteryOptimization.isExempt(context) }.distinctUntilChanged()`, driven from `refresher.ticks` (the same `NotificationPermissionRefresher` that fires ON_RESUME). `batteryExemptStarted` wraps with `onStart { emit(true) }` to avoid an initialization flash. Reactive-clear test confirms the banner drops when the flow flips.

**3. Dialog action + `package:` URI + manifest** — `requestExemptionIntent` correctly uses `Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS` + `Uri.parse("package:${context.packageName}")`. `REQUEST_IGNORE_BATTERY_OPTIMIZATIONS` declared in the shared manifest with a clear comment on the Play policy rationale. Applied to **both** the new Home banner (`MainDashboardDestination`) and the existing #453 Settings "Fix this" (`SettingsDestination`). `startActivitySafely` promoted to `internal` for the cross-file use — appropriate for an app module.

**4. Tests** — `BatteryOptimizationTest` verifies action + `package:` data URI. `MainDashboardViewModelTest` covers all four required scenarios: shows on foss/not-exempt, absent on foss/exempt, absent on google/not-exempt, and reactive-clear (MutableStateFlow flip). Tests are meaningful — the first two would fail without the banner logic.

**UX decisions doc** — Entry exists, dated 2026-06-16, describes both parts, explicitly supersedes the 2026-06-02 #453 "Fix this opens battery-optimization settings" decision.

**Quality** — FQN for `LocalExtendedColors` is consistent with the pre-existing `DeliveryBannerCard`. No GlobalScope, structured concurrency respected, `combine` arity correct, coroutine rules clean. ktlint/detekt/Build/test CI all green (confirmed remote).

Issue #389, opened 2026-04-23
Fresh install doesn't provision certs — OnboardingFlow doesn't call CertProvisioningFlow (root cause of #386 + #387)


Monkopedia:
Root cause of both #386 and #387, found by tracing the cert-provisioning flow.

## The bug

`CertProvisioningFlow.execute()` — the method that calls the relay's `/registerCerts` endpoint and saves the three PEMs (`rouse_cert.pem` server cert, `rouse_client_cert.pem` mTLS client cert, `rouse_relay_ca.pem`) — is only invoked by `IntegrationSetupViewModel`, when the user explicitly adds their first integration.

`OnboardingFlow.execute()` (the flow driven by the Welcome screen) registers the device with the relay, gets a subdomain, and stores integration secrets. It does NOT provision certs.

So after a fresh install with onboarding-only (no integrations added):
- `rouse_subdomain.txt` ✓
- `rouse_integration_secrets.json` ✓
- `rouse_cert.pem` ✗ (missing)
- `rouse_client_cert.pem` ✗ (missing)
- `rouse_relay_ca.pem` ✗ (missing)

The device is in a "half-onboarded" state that appears functional (subdomain listed, integrations in Pending state, dashboard renders) but is structurally broken:

- **AI client → device (server role):** no `rouse_cert.pem` → `CertStoreTlsCertProvider.serverSslContext()` returns null → `SessionHandler.handleStream()` silently fails → client sees EOF. This is #386.
- **Device → relay reconnect (client role):** no `rouse_client_cert.pem` → WS mTLS upgrade gets 401. This is #387.

Initial WS connect succeeds because the first WS upgrade uses the Firebase ID token for auth (before mTLS kicks in), which is why the Pixel's tunnel came up briefly at 21:49 UTC and then failed to reconnect at 21:54.

## Evidence (from Pixel 6 Pro on adolin, fresh install at 2026-04-23 16:55)

Files on disk after onboarding but no integration enabled:

```
$ run-as com.rousecontext.debug ls files/
rouse_integration_secrets.json
rouse_subdomain.txt      # "rusty"
datastore/
.crashlytics.v3/
generatefid.lock
profileInstalled
```

No `rouse_cert.pem`, no `rouse_client_cert.pem`, no `rouse_relay_ca.pem`.

[...]

PR #489, opened 2026-06-16
Decouple foss wake-activation from cert provisioning (#486)

### LINKED ISSUE #486: FOSS: Home shows 'On-demand wake is off' immediately after completing delivery setup in onboarding (must redo to clear)
## Bug
On a fresh FOSS install, after completing onboarding **including** the Background delivery setup (picked a distributor), Home still showed the **"On-demand wake is off"** degraded banner — despite delivery having just been set up. The user had to go through delivery setup **again** for it to take. So a correctly-onboarded foss user lands on a degraded Home and is told setup failed when it (mostly) didn't.

## Repro
Fresh foss install (cleared data) → onboarding → pick distributor on Background delivery → land on Home → "On-demand wake is off" banner present. Re-run delivery setup → banner clears, wake works. (Observed live on the user's Pixel 10, 2026-06-16.)

## Two candidate root causes (needs investigation to disambiguate — do NOT assume)
1. **#480 stopped-state race:** the freshly-installed distributor (ntfy) hadn't been launched, so its UnifiedPush `REGISTER`/`NEW_ENDPOINT` never fired during onboarding → no endpoint → genuinely-degraded → relaunching ntfy (+ redo) delivered the endpoint. If so, this is a manifestation of #480, and the fix is the #480 nudge (launch the distributor / wait-and-retry for the endpoint).
2. **Home-banner refresh race:** the endpoint/registration *did* arrive but Home evaluated the delivery-activation state before it propagated, showing a stale degraded banner that never re-evaluated until the user forced it. If so, it's a state-observation bug in the Home delivery-banner (the banner should reflect the registration completing without a manual redo).

## Notes
- FOSS-only (google/FCM has no distributor step).
- Cross-refs #480 (#480 may be the cause or a sibling). Determine which before fixing — the fixes differ.
- Could be investigated live: relay logs (when did the endpoint/registration land for this device vs when Home showed degraded) + the foss delivery-activation/Home-banner state flow.

monkopedia-coder:
## Problem

On a fresh foss onboard, `UnifiedPushBackgroundDelivery.registerWithEndpoint` latched `DeliveryActivation.NeedsSetup` for ANY non-`Success` result of `OnboardingFlow.execute`. But `execute` registers + persists the subdomain FIRST, then chains ACME cert provisioning (#389). A cert-stage failure (`CertRateLimited` / `CertRelayError` / `CertNetworkError` / `CertKeyGenerationFailed` / `CertStorageFailed`) returns non-`Success` while the device is **already registered with its push endpoint reported**.

Those results hit the `else` branch and wrongly latched `NeedsSetup`, so Home showed "On-demand wake is off / set up a delivery app" even though wake was fully configured and only the cert hop stumbled. (A redo cleared it because the new endpoint routed through `refreshEndpoint`, which sets `Active` unconditionally.) Cert problems are already surfaced + retried independently via the dashboard cert-renewal banner, so they must not flip the delivery banner.

## Fix

Key wake-activation off the persisted subdomain (mirroring the `init` seeding and `refreshEndpoint`) rather than full-flow success:
- `Active` when `certificateStore.getSubdomain() != null` (registered with a wake target),
- `NeedsSetup` only for pre-subdomain failures (no subdomain persisted),
- still `registrationStatus.markComplete()` on `Success`.

foss-only; no change to the cert-renewal banner, ViewModel, or Home.

## Tests

Adds the first direct unit test for `UnifiedPushBackgroundDelivery` (`app/src/testFoss/.../UnifiedPushBackgroundDeliveryTest.kt`):
- `Cert*` failure with a persisted subdomain -> `Active` (red against old code, green now),
- pre-subdomain failure (no subdomain) -> `NeedsSetup`,
- `Success` -> `Active` + `markComplete()`.

## Verification

- `ktlintCheck` + `detekt` clean

Closes #486

### DIFF
--- a/app/src/foss/java/com/rousecontext/app/delivery/UnifiedPushBackgroundDelivery.kt
+++ b/app/src/foss/java/com/rousecontext/app/delivery/UnifiedPushBackgroundDelivery.kt
@@ -152,23 +152,29 @@ class UnifiedPushBackgroundDelivery(
             Log.e(TAG, "Couldn't obtain device credential for deferred registration", e)
             null
         } ?: return
-        when (
-            val result = onboardingFlow.execute(
-                credential,
-                fcmToken = "",
-                pushEndpoint = endpoint
-            )
-        ) {
-            is OnboardingResult.Success -> {
-                registrationStatus.markComplete()
-                _activation.value = DeliveryActivation.Active
-                Log.i(TAG, "Deferred foss registration complete; on-demand wake active")
-            }
-            else -> {
-                Log.w(TAG, "Deferred foss registration did not complete: $result")
-                _activation.value = DeliveryActivation.NeedsSetup
-            }
+        val result = onboardingFlow.execute(
+            credential,
+            fcmToken = "",
+            pushEndpoint = endpoint
+        )
+        if (result is OnboardingResult.Success) {
+            registrationStatus.markComplete()
         }
+        // Wake-activation tracks "is there a registered push wake target?", which
+        // is exactly "did we persist a subdomain?" — NOT "did the whole flow incl.
+        // certs succeed" (#486). `execute` registers + persists the subdomain
+        // before chaining ACME cert provisioning (#389), so a cert-stage failure
+        // (Cert*) returns non-Success while the device is already registered with
+        // its endpoint reported. Keying off the persisted subdomain (mirroring the
+        // init seeding and refreshEndpoint) keeps the delivery banner Active in
+        // that case; cert problems are surfaced separately by the cert-renewal
+        // banner. Only pre-subdomain failures leave wake needing setup.
+        _activation.value = if (certificateStore.getSubdomain() != null) {
+            DeliveryActivation.Active
+        } else {
+            DeliveryActivation.NeedsSetup
+        }
+        Log.i(TAG, "Deferred foss registration result=$result; activation=${_activation.value}")
     }
 
     private suspend fun refreshEndpoint(endpoint: String) {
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Reviewed the subdomain-presence discriminator, pre-subdomain-failure path, test meaningfulness, and scope. All four points clear.

**Correctness**

1. **Subdomain-presence discriminator is correct.** `OnboardingFlow.execute(credential, fcmToken, pushEndpoint)` calls `registerAndPersist`, which calls `certificateStore.storeSubdomain()` *before* returning `RegisterOutcome.Success`; only after that does `provisionCerts` run, which is the sole source of all `Cert*` variants. So `getSubdomain() != null` is exactly right as the discriminator — `Cert*` failures always land with a persisted subdomain.

2. **Pre-subdomain failures yield NeedsSetup.** Registration failures (`RateLimited`, `RelayError`, `NetworkError`) return from `registerAndPersist` before `storeSubdomain` is called; `StorageFailed` calls `certificateStore.clear()` before returning. All four leave `getSubdomain() == null`, so the fix correctly yields `NeedsSetup`.

3. **Tests are meaningful.** The two `Cert*`-with-subdomain cases (`CertRateLimited`, `CertRelayError`) would have been red against the old `else -> NeedsSetup` branch and are green now. The mock correctly mirrors the real register-before-certs ordering by persisting `persistedSubdomain` inside `coAnswers` before returning the cert-failure result. The pre-subdomain and Success cases round out regression coverage. `runTest` + `advanceUntilIdle()` is correct here (no real I/O; the delivery is constructed with the test scope so all launched coroutines are drained by `advanceUntilIdle`).

4. **No scope creep.** Only two files changed: the production foss delivery (38-line targeted delta) and the new test. Cert-renewal banner, ViewModel, and Home are untouched.

CI: all 3 checks pass (Build & Test, Lint, test).

Issue #487, opened 2026-06-16
Notification-listener access: deep-link to our listener's toggle (ACTION_NOTIFICATION_LISTENER_DETAIL_SETTINGS) instead of the full list


monkopedia-coder:
## Improvement (from the 2026-06-16 permission-UX audit)
For the notification-capture integration, enabling notification-listener access currently opens **`ACTION_NOTIFICATION_LISTENER_SETTINGS`** — the full list of every notification-listener app, where the user has to find ours (`NotificationSetupDestination.kt:78`).

On **API 30+** we can use **`ACTION_NOTIFICATION_LISTENER_DETAIL_SETTINGS`** + `EXTRA_NOTIFICATION_LISTENER_COMPONENT_NAME` (our listener's `ComponentName`) to deep-link **straight to our listener's own enable toggle**. Fall back to the current list action on API < 30.

## Change
- `NotificationSetupDestination.kt:78` — branch on `Build.VERSION.SDK_INT >= 30`: detail action + component-name extra; else the existing list action.
- No new manifest permission. Same onboarding step, same intent purpose — only the deep-link target tightens.

## Notes
- minSdk 24, so the API<30 fallback is required.
- **Gray-zone UX:** same step / same intent, only the target tightens (list → our toggle). Likely fine as a targeting tweak rather than a flow change, but worth a quick confirm; if treated as an error-surface change, add a `docs/ux-decisions.md` note in the PR.
- Audit also confirmed everything else (POST_NOTIFICATIONS runtime dialog, app-notification-settings recovery link, Health Connect native contract, overlay targeted deep-link) is already best-available; DND + Usage-access have no better standard mechanism than the list.

PR #492, opened 2026-06-16
Nudge user to open a freshly-installed UnifiedPush distributor (#480)

### LINKED ISSUE #480: FOSS: freshly-installed UnifiedPush distributor stays in 'stopped' state until launched once (device stays degraded)
## Observed (on-device, 2026-06-14, adolin)
On a fresh FOSS setup, a freshly-**installed** ntfy distributor does not receive the UnifiedPush `REGISTER` broadcast until ntfy has been **launched at least once** (Android holds newly-installed apps in a "stopped" state that suppresses broadcasts until first launch). Result: the user picks ntfy in the Background delivery picker, but the device stays **degraded/unregistered** — no endpoint arrives — until they manually open ntfy once. From the user's view it looks like setup silently failed.

## Why it needs a decision
There are 2+ defensible UX shapes with different trade-offs, so this wants a direction call before implementation:
- **Detect + nudge:** after the user selects a just-installed distributor, if no `NEW_ENDPOINT` arrives within a timeout, surface "Open ntfy once to finish setup" (with a launch button).
- **Proactively launch the distributor:** fire a launch intent for the chosen distributor right after selection to clear the stopped state. (More magical; may surprise the user / not all distributors like being launched.)
- **Document-only:** add a one-liner to the picker copy ("you may need to open the app once").

## Notes
- Specific to the FOSS/UnifiedPush flow (#463); google/FCM is unaffected.
- UX-affecting (onboarding/first-run + error surface) → needs in-session approval + a `docs/ux-decisions.md` entry per `.claude/rules/ux-changes.md`. Filing as `needs-decision`.
- Part of #460.

monkopedia-coder:
## What

A freshly-installed UnifiedPush distributor (e.g. ntfy) sits in Android's "stopped" state and silently drops broadcasts — including our UnifiedPush `REGISTER` — until it's launched once. So a user could pick a distributor in onboarding and have **no endpoint ever arrive**, leaving the device degraded with no signal.

This implements the approved **option (a) detect-and-nudge**: after the user selects a distributor, the picker waits briefly for the push endpoint. If it doesn't arrive (the stopped-state case), the picker shows a dismissible nudge whose action button **launches the chosen distributor** (clearing the stopped state so its registration fires). The endpoint then flows back, the nudge clears automatically, and the flow advances.

Copy: **"Open {distributor} to enable"** — the chosen distributor's display name is interpolated (never hardcoded to ntfy).

## Changes

- `BackgroundDelivery`: new `endpointArrived: StateFlow<Boolean>` signal + `launchIntent(id)`. Foss wires them to the persisted/arriving endpoint and `packageManager.getLaunchIntentForPackage`; `NoOpBackgroundDelivery` returns `false`/`null`.
- `BackgroundDeliveryViewModel`: after `select`, wait-with-timeout for the endpoint; raise/clear a `DistributorNudge`; emit `proceed` once the endpoint arrives.
- `BackgroundDeliveryScreen`: render the nudge card with the "Open {distributor} to enable" action.
- `BackgroundDeliveryDestination`: defer the advance until `proceed` (instead of firing on tap); launch the distributor on the nudge action.
- `docs/ux-decisions.md`: entry for the approved decision.

## Tests

- `BackgroundDeliveryViewModelTest`: endpoint-arrives-promptly → no nudge + proceeds; no-endpoint-within-timeout → nudge naming the distributor; endpoint-arrives-after-nudge → clears + proceeds; dismiss hides it.
- `./gradlew :app:testDebugUnitTest` (foss/default) green; `:app:compileDebugKotlin -Pgoogle` compiles; `:app:ktlintCheck detekt` clean.

Closes #480


monkopedia-reviewer (approved):
Reviewed against focus points and coroutine rules. All green.

**Structured concurrency:** `viewModelScope.launch` + `withTimeoutOrNull` — no GlobalScope, no runBlocking in production code. `selectionJob?.cancel()` on re-select. ViewModel scope teardown covers the job automatically.

**No deadlock:** After the timeout the job stays alive blocking on `delivery.endpointArrived.first { it }`. When the endpoint arrives (even post-dismiss), `_nudge.value = null` then `_proceed.emit(Unit)` → `LaunchedEffect` calls `onChosen()`. The `proceed` SharedFlow carries `extraBufferCapacity = 1`, so an early emit (endpoint arrives before the composable is collected) is buffered and received correctly. A healthy distributor auto-advances without ever raising the nudge.

**Navigation timing:** advance-on-event is sound. `LaunchedEffect(Unit)` begins collecting `proceed` on composition; the buffer handles the race window on fast endpoints. `DistributorOption.Kind.ACTIVE` path no longer calls `onChosen()` directly — only `proceed` advances the flow. ux-decisions.md entry covers the trade-off.

**Name interpolation + null launchIntent:** distributor name is `_rows.value.firstOrNull { it.id == id }?.name ?: id`; copy never hardcodes "ntfy". `launchIntent` nullable handled via `.let` + `ActivityNotFoundException` catch.

**Tests:** cover all four cases (prompt-arrives / timeout-nudge / arrives-after-nudge / dismiss) with `StandardTestDispatcher + advanceUntilIdle`. `FakeDelivery` in both test files updated for the new interface members. `NoOpBackgroundDelivery` returns `false`/`null` — Google path unaffected.

**Merge resolution:** `docs/ux-decisions.md` on the PR branch has both the #480 entry (top) and the #487 entry intact — no content lost.

**CI:** all 3 checks passed (Build & Test 9m47s, Lint 2m24s, test 28m31s).

PR #494, opened 2026-06-16
Re-register distributor after opening it from the #480 nudge

### LINKED ISSUE #493: FOSS #480 nudge: opening the distributor doesn't re-register, so the nudge never clears (dropped REGISTER not re-issued)
## Bug (found in on-device validation of the merged #480 fix — #492)
The #480 "Open {distributor} to enable" nudge appears correctly when a fresh/stopped distributor is selected, and tapping it **launches** the distributor — but the **nudge never clears and the endpoint never registers**, so the picker is stuck until the user manually re-selects the distributor row.

## Repro (adolin, foss debug @ main 9fb6f785)
1. Install ntfy fresh, don't launch it (Android "stopped" state); fresh-onboard the app; select ntfy.
2. Nudge appears after the 6s timeout ("We haven't heard from ntfy yet" + "Open ntfy to enable"). ✓
3. Tap the nudge button → ntfy launches (stopped→running). ✓
4. **Nudge never clears; no `REGISTER`/`NEW_ENDPOINT` ever arrives** (waited 40s+). Manually re-tapping the ntfy row (re-invokes `registerApp`) immediately produces `REGISTER received` → `NEW_ENDPOINT` → nudge clears → advances.

## Root cause
The fix's premise — "launching the distributor clears the stopped state so its registration fires" — is incomplete: there is **no pending registration to fire.** `selectDistributor()` calls `UnifiedPush.registerApp()` exactly once at selection time; that REGISTER broadcast is **dropped** because the distributor is stopped. UnifiedPush does **not** redeliver it. The nudge action `onOpenDistributor` only does `startActivity(launchIntent(id))` — it launches the distributor but never **re-issues `registerApp`**. On return, the destination's `ON_RESUME` only calls `rescan()`, not a re-register. So the now-running distributor never receives a REGISTER, never mints an endpoint, `endpointArrived` never flips, the nudge never clears, `proceed` never fires.

## Fix
After launching the distributor (in `onOpenDistributor`), and/or on `ON_RESUME` when there's a selected-but-not-yet-acked distributor, **re-call `registerApp`/`select`** so the REGISTER fires now that the distributor is un-stopped. (Re-selecting the row already works — wire that same path into the nudge action / resume.)

Relevant source:
- `app/src/main/java/com/rousecontext/app/ui/navigation/destinations/BackgroundDeliveryDestination.kt` — `onOpenDistributor` launches only; `ON_RESUME` → `rescan()` only.
- `app/src/main/java/com/rousecontext/app/ui/viewmodels/BackgroundDeliveryViewModel.kt` — `select()` registers once; the nudge path awaits `endpointArrived` forever.
- `app/src/foss/java/com/rousecontext/app/delivery/UnifiedPushBackgroundDelivery.kt` — `selectDistributor` = save + `registerApp`; `launchIntent` is launch-only.

## Add a test
Extend `BackgroundDeliveryViewModelTest`: after the nudge is shown, invoking the open/re-register path must re-issue registration so `endpointArrived` can flip and `proceed` fires.

## Notes
- Regression in the just-merged #480 (#492). The nudge UI/copy/timeout all work; only the re-register-on-open step is missing.
- Adjacent server behavior observed (NOT a defect to fix here): a freshly-installed ntfy whose `SubscriberService` hasn't connected yet causes ntfy.sh to reject relay wakes with `507 "cannot publish to UnifiedPush topic without previously active subscriber"` until ntfy is opened and its foreground subscriber connects. Opening ntfy (which the nudge is meant to drive) also establishes that subscriber — another reason the nudge must actually complete.
- Found by the on-device validation sweep; the other 5 items (onboarding, #483 dialog, #487 deep-link, #484 switch, end-to-end UnifiedPush wake) PASSED.

monkopedia-coder:
Closes #493

## Problem
The #480 "Open {distributor} to enable" nudge appeared and launched the distributor, but never cleared and no endpoint ever arrived. `select()` issues a single UnifiedPush `REGISTER`, which Android drops while the freshly-installed distributor is in its stopped state; UnifiedPush does not redeliver it. The nudge's "Open" action only launched the distributor — it never re-issued `REGISTER` — so the now-running distributor had nothing to act on. `ON_RESUME` only called `rescan()`. Net result: no endpoint, `endpointArrived` never flipped, the nudge stuck forever (only a manual re-select recovered).

## Fix
Re-issue registration once the distributor has come up, reusing the known-good `selectDistributor` path (manual re-select already worked):

- `BackgroundDeliveryViewModel.onDistributorOpened(id)` — launched by the nudge's "Open" action after the destination starts the distributor. Waits a short, cancellable `~1.5s` (`REREGISTER_DELAY_MS`) for the distributor to leave the stopped state, then re-registers. Runs in `viewModelScope`; a `reregisterJob` is cancelled/replaced on repeat taps. The in-flight `selectionJob` is still awaiting `endpointArrived`, so when the endpoint arrives the nudge clears and `proceed` fires.
- `BackgroundDeliveryViewModel.onResume()` — destination `ON_RESUME` now calls this (was `rescan()`). Re-scans and, if a distributor was selected but no endpoint has arrived yet, re-registers — belt-and-suspenders for users who linger in the distributor app past the delay. No-op once the endpoint is in hand.
- Nudge remains a non-permanent fallback: if the endpoint still doesn't arrive, the nudge stays up so the user can retry; it never gets stuck.

No new interface surface — both paths reuse `BackgroundDelivery.selectDistributor`, which is exactly what manual re-selection does.

## Tests
Extended `BackgroundDeliveryViewModelTest` with a fake whose `selectDistributor` only delivers an endpoint on the **second** call (simulating the dropped-while-stopped first REGISTER):
- `opening the distributor re-registers so a dropped REGISTER is re-issued` — nudge shown after first (dropped) register; `onDistributorOpened` re-registers, endpoint arrives, nudge clears, `proceed` fires.
- `ON_RESUME re-registers a selected distributor that has no endpoint yet` — same via `onResume()`.
- `ON_RESUME does not re-register once the endpoint has arrived` — guard verified (no extra register).

Both new "re-register" tests fail against the pre-fix code (nudge never clears) and pass after.

## Verification
- `ktlintCheck detekt` clean

### DIFF
--- a/app/src/main/java/com/rousecontext/app/ui/navigation/destinations/BackgroundDeliveryDestination.kt
+++ b/app/src/main/java/com/rousecontext/app/ui/navigation/destinations/BackgroundDeliveryDestination.kt
@@ -73,12 +73,14 @@ private fun BackgroundDeliveryDestinationContent(
     val pendingIntegration by pendingSetup.pendingId.collectAsState()
 
     // Re-scan installed distributors when the user returns (e.g. after
-    // installing one from the store the "install" rows deep-link to).
+    // installing one from the store the "install" rows deep-link to). #493: also
+    // re-register a selected-but-unacked distributor, in case the user opened it
+    // from the nudge and lingered there past the VM's re-register delay.
     val lifecycleOwner = LocalLifecycleOwner.current
     DisposableEffect(lifecycleOwner.lifecycle) {
         val observer = LifecycleEventObserver { _, event ->
             if (event == Lifecycle.Event.ON_RESUME) {
-                viewModel.rescan()
+                viewModel.onResume()
             }
         }
         lifecycleOwner.lifecycle.addObserver(observer)
@@ -153,6 +155,10 @@ private fun BackgroundDeliveryDestinationContent(
             delivery.launchIntent(distributorNudge.distributorId)?.let { intent ->
                 try {
                     context.startActivity(intent)
+                    // #493: launching alone won't register — the original REGISTER
+                    // was dropped while the distributor was stopped. Re-issue it
+                    // once the distributor has had a moment to come up.
+                    viewModel.onDistributorOpened(distributorNudge.distributorId)
                 } catch (_: ActivityNotFoundException) {
                     // Distributor has no launchable entry; leave the nudge up.
                 }
--- a/app/src/main/java/com/rousecontext/app/ui/viewmodels/BackgroundDeliveryViewModel.kt
+++ b/app/src/main/java/com/rousecontext/app/ui/viewmodels/BackgroundDeliveryViewModel.kt
@@ -5,6 +5,7 @@ import androidx.lifecycle.viewModelScope
 import com.rousecontext.app.delivery.BackgroundDelivery
 import com.rousecontext.app.delivery.DistributorOption
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharedFlow
@@ -32,6 +33,15 @@ import kotlinx.coroutines.withTimeoutOrNull
  * clears the stopped state). When the endpoint finally arrives the nudge clears
  * and the picker advances via [proceed]. If the endpoint arrives promptly there
  * is no nudge and the picker advances immediately.
+ *
+ * ## Re-register after opening (issue #493)
+ *
+ * Launching the distributor is not enough on its own: the `REGISTER` from
+ * [select] was already dropped while the distributor was stopped, and UnifiedPush
+ * never redelivers it, so the now-running distributor has nothing to act on. The
+ * nudge's "Open" action therefore re-issues registration shortly after launch
+ * ([onDistributorOpened]), and the destination's `ON_RESUME` ([onResume]) does
+ * the same if the user returns with a selected-but-unacked distributor.
  */
 class BackgroundDeliveryViewModel(private val delivery: BackgroundDelivery) : ViewModel() {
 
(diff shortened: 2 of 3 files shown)


monkopedia-reviewer (approved):
Clean fix for the #480 nudge hang. All 6 focus points verified:

1. **Structured/cancellable coroutine** — `onDistributorOpened` uses `viewModelScope.launch` with `reregisterJob?.cancel()` before replacing; `delay()` is a cancellable suspension point. No `GlobalScope` or `runBlocking` in production code.

2. **Hang genuinely fixed** — `selectionJob` from `select()` is still alive awaiting `endpointArrived`; when `onDistributorOpened` re-issues `selectDistributor` after the delay and the endpoint arrives, `selectionJob` sees it, clears the nudge, and fires `proceed`. Confirmed by the first new test.

3. **Non-permanent fallback** — if the re-register still doesn't produce an endpoint, the nudge stays visible (the timeout already fired; `selectionJob` continues waiting indefinitely); the user can tap "Open" again, which cancels and replaces `reregisterJob`.

4. **`onResume` is a no-op once endpoint is in hand** — `onResume()` guards with `!delivery.endpointArrived.value` before calling `selectDistributor`. Third test verifies: `delivery.selected.size` stays at 1 after `onResume()` when endpoint already arrived.

5. **Google path unaffected** — `NoOpBackgroundDelivery.selectDistributor` is not touched; both new VM paths call through the same `delivery.selectDistributor` interface.

6. **Tests exercise the regression with test scheduler** — `deliverEndpointOnSelectCall = 2` drops the first REGISTER (simulating stopped-state); new tests use `runTest(dispatcher)` + `advanceUntilIdle()` to advance through the 1.5 s delay. Both regression tests would be red against pre-fix code.

CI: Build & Test ✓, Lint ✓, test ✓

PR #499, opened 2026-06-18
Wire real CertInspector and reach the CN-mismatch path (#495)

### LINKED ISSUE #495: CertInspector is a no-op stub in production — CN-mismatch and in-flow expiry checks never fire
**Category:** dead-code / correctness-adjacent (a wired-up safety check is inert in production)
**Where:**
- `core/tunnel/src/jvmMain/kotlin/com/rousecontext/tunnel/CertRenewalFlow.kt:253-257` — the `CertInspector` base class and `CertInfo`
- consumers in the same file: the `isExpired` guard at `:85` and the CN-mismatch check at `:163-174`
- production wiring: `app/src/main/java/com/rousecontext/app/di/AppModule.kt:264` constructs `CertRenewalFlow(...)` **without** a `certInspector` argument

**Current:** `CertInspector` is an `open class` whose default `inspect()` returns an empty stub:

```kotlin
open class CertInspector {
    open fun inspect(pemCertificate: String): CertInfo = CertInfo() // commonName=null, isExpired=false
}
```

The only override anywhere in the tree is in `CertRenewalTest.kt`. Production (`AppModule.kt:264`) omits the `certInspector` parameter, so it gets the default stub. Consequences in production:
- `renewWithMtls` line 85 — `if (certInspector.inspect(currentCert).isExpired) return RenewalResult.CertExpired` — `isExpired` is always `false`, so this guard never fires. (`CertRenewalWorker.kt:192` even has handling for the `CertExpired` result with a log "mTLS renewal rejected: cert reported as expired" that can never be reached this way.)
- `handleRenewResponse` lines 163-174 — the CN-mismatch defense compares `oldInspection.commonName` to `newInspection.commonName`, but both are always `null`, so `null != null` is `false` and `RenewalResult.CnMismatch` is never produced. Yet there is a full consumer chain for it: `CertRenewalWorker.handleCnMismatch` (`:221/:246`) → `CertRenewalBannerFlow` → a user-facing `TerminalReason.CnMismatch` banner. The detection that drives that entire path is dead in production.

(The worker's real expiry decision uses `store.getCertExpiry()`, not the inspector, so renewal scheduling itself is unaffected — but the in-flow guard and the CN-mismatch safety net are both no-ops.)

**Proposed:** Pick one direction (staged `needs-decision`, not auto-worked):
- **(a) Wire a real inspector** — provide an X.509-backed `CertInspector` (parse the PEM for CN + notAfter; the app already parses certs in `app/.../cert/`) and inject it at `AppModule.kt:264`, so the CN-mismatch safety check and the in-flow expiry guard actually work. Add a ViewModel/flow test that a CN change yields `CnMismatch`.
- **(b) Remove the dead machinery** — if the CN-mismatch defense is not wanted, delete `CertInspector`/`CertInfo`, the `isExpired`/CN checks, the `RenewalResult.CnMismatch` arm, and the worker's `handleCnMismatch` path (note: this trims the excluded Home banner's `TerminalReason.CnMismatch` case too, so coordinate).

**Why:** Right now there is an entire user-facing error path (CN-mismatch banner) hanging off a check that can never trigger, plus an unreachable `CertExpired` branch. Either make the protection real or remove the dead weight — both eliminate a misleading "this is handled" surface.
**Effort:** small refactor (b) / larger (a — real X.509 inspector + tests)

monkopedia-coder:
Closes #495

## What
The certificate-renewal CN-mismatch safety net and the expiry guard in `CertRenewalFlow` were both inert in production:
- `CertInspector` was a no-op `open class` stub returning an empty `CertInfo()`, so `isExpired` was always `false` and the old-vs-new CN comparison was always `null == null`.
- As a result the whole downstream banner chain (`RenewalResult.CnMismatch` -> `CertRenewalWorker.handleCnMismatch` -> `CertRenewalBannerFlow` -> dashboard banner) could never fire, even though it is fully built and tested.

## Changes
1. **Real `CertInspector.inspect(pem)`** — parses the PEM with `CertificateFactory` into an `X509Certificate`, populating `commonName` (subject CN via `LdapName`) and `isExpired` (`notAfter < now`). The class stays `open` so existing test fakes still override it. A malformed/unparseable cert degrades safely to an empty `CertInfo()` rather than throwing inside the renewal flow.
2. **CN-mismatch path made reachable on every renewal path** — `renewWithFirebase` and `renewExpired` (keypair + Firebase fallback branches) now pass the current stored cert into `handleRenewResponse` instead of `null`. The mTLS path already passed it. The expired current cert still carries a valid subject CN, so the comparison is meaningful there too.

No production wiring change at the injection site is needed: `AppModule` already constructs `CertRenewalFlow` with the default `CertInspector`, which is now real.

## Tests
Extended `CertRenewalTest` with real openssl-generated self-signed fixtures (so CN + expiry extraction runs against the production code path):
- real `CertInspector` parses CN + reports non-expired for a valid cert
- real `CertInspector` flags an expired cert
- real `CertInspector` degrades to empty `CertInfo` on malformed input
- mTLS CN-mismatch fires end-to-end with the real inspector
- Firebase-path CN-mismatch fires end-to-end (guards the new currentCert wiring)
- expired current cert returns `CertExpired` with the real inspector

All 5 were red before the implementation and green after. Existing tests (including the override-based CN-mismatch test) are unchanged.

## Verification
- `:core:tunnel:jvmTest` green
- `:core:tunnel:ktlintCheck` + whole-tree `detekt` clean

### DIFF
--- a/core/tunnel/src/jvmMain/kotlin/com/rousecontext/tunnel/CertRenewalFlow.kt
+++ b/core/tunnel/src/jvmMain/kotlin/com/rousecontext/tunnel/CertRenewalFlow.kt
@@ -1,5 +1,10 @@
 package com.rousecontext.tunnel
 
+import java.io.ByteArrayInputStream
+import java.security.cert.CertificateFactory
+import java.security.cert.X509Certificate
+import java.time.Instant
+import javax.naming.ldap.LdapName
 import kotlinx.coroutines.delay
 
 /**
@@ -145,6 +150,9 @@ class CertRenewalFlow(
     ): RenewalResult {
         val subdomain = certificateStore.getSubdomain()
             ?: return RenewalResult.NoCertificate
+        // The current (typically expired) cert still carries a valid subject CN, so pass it
+        // into validation: the CN-mismatch safety net must fire on the expired-cert path too.
+        val currentCert = certificateStore.getCertificate()
 
         val csrResult = try {
             val keyPair = deviceKeyManager.getOrCreateKeyPair()
@@ -163,7 +171,7 @@ class CertRenewalFlow(
                 firebaseToken = credentials.token,
                 signature = credentials.signature
             )
-            handleRenewResponse(result, null)
+            handleRenewResponse(result, currentCert)
         }
     }
 
@@ -183,6 +191,9 @@ class CertRenewalFlow(
     ): RenewalResult {
         val subdomain = certificateStore.getSubdomain()
             ?: return RenewalResult.NoCertificate
+        // The current (expired) cert still carries a valid subject CN; pass it into validation
+        // so the CN-mismatch safety net runs on both the keypair and Firebase fallback branches.
+        val currentCert = certificateStore.getCertificate()
 
         val csrResult = try {
             val keyPair = deviceKeyManager.getOrCreateKeyPair()
@@ -199,7 +210,7 @@ class CertRenewalFlow(
                     csrSignature = keypair.csrSignature,
                     proof = keypair.proof
                 )
-                handleRenewResponse(result, null)
+                handleRenewResponse(result, currentCert)
             }
         }
 
@@ -213,7 +224,7 @@ class CertRenewalFlow(
                 firebaseToken = credentials.token,
                 signature = credentials.signature
             )
-            handleRenewResponse(result, null)
+            handleRenewResponse(result, currentCert)
         }
     }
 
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (changes requested):
## Tier-1 Blocker: `javax.naming.ldap.LdapName` is not available on Android

### Verdict: request_changes

The real `CertInspector` uses `javax.naming.ldap.LdapName` to parse the subject CN. **`javax.naming` is absent from the Android API surface.** I verified this directly against the android.jar stubs:

```
jar tf Android/Sdk/platforms/android-35/android.jar | grep 'javax/'
# -> javax/annotation, javax/crypto, javax/microedition, javax/net, javax/security, javax/sql, javax/xml
# NO javax/naming
```

The `desugar_jdk_libs 2.1.5` JAR (the project's core library desugaring artifact) also contains no `javax/naming` entries. There is no path for `LdapName` to reach the Android runtime.

### Why this doesn't crash — but still breaks the feature

`extractCommonName` wraps the `LdapName` call in `runCatching { ... }.getOrNull()`. In Kotlin, `runCatching` catches **`Throwable`** — including `Error` subclasses such as `NoClassDefFoundError`. So on Android:

1. `LdapName(...)` throws `NoClassDefFoundError` at runtime.
2. `runCatching` silently swallows it.
3. `extractCommonName` always returns `null` on Android.
4. `CertInfo.commonName` is always `null`.
5. The CN-mismatch check becomes `null != null` → `false` → **mismatch never detected on Android**.

The PR's primary security goal — making the CN-mismatch safety net actually fire — remains a permanent no-op on Android, the production platform, despite the JVM tests passing. The expiry check still works (it uses `java.security.cert.X509Certificate` + `java.time.Instant` which are both present), but the CN guard is dead.

### Why `jvmTest` can't catch this

`core/tunnel` is a KMP module with only a `jvm()` target. All tests run on JDK, where `javax.naming` is available. The bug only manifests on the Android ART runtime, which `jvmTest` never exercises.

### Required fix

Replace the `LdapName` call with a direct parse of `X500Principal.getName()` (RFC 2253 format, e.g. `"CN=foo.example,O=..."`). No JNDI required — a regex or simple split works:

```kotlin
private fun extractCommonName(cert: X509Certificate): String? = runCatching {
    cert.subjectX500Principal.name  // e.g. "CN=foo.example.com,O=...,C=..."
        .split(',')
        .map { it.trim() }
        .firstOrNull { it.startsWith("CN=", ignoreCase = true) }
        ?.removePrefix("CN=")
        ?.removePrefix("cn=")  // RFC 2253 can be mixed case
}.getOrNull()
```

Or use a Regex: `Regex("(?:^|,)\\s*CN=([^,]+)", RegexOption.IGNORE_CASE)` against `subjectX500Principal.name`. Either eliminates the JNDI dependency and is safe on Android API 24+.

The `remove javax.naming.ldap.LdapName import` must also be removed from the file or the JVM build will fail.

---
*Reviewed by monkopedia-reviewer. This PR is owner-gated — do not merge.*

monkopedia-reviewer (approved):
Re-review of LdapName fix (issue #495/#499). Prior security review cleared + owner-approved merge. Fix confirmed: CN now parsed via X500Principal.name RFC 2253 string with hand-written splitRfc2253 — no javax.naming/JNDI, Android-safe. Degrades safely on parse failure. Tests use real parsed ECDSA certs and explicitly guard the JNDI-free path. App-side only (core/tunnel, no relay changes). CI green (Build, Lint, test all passed).

PR #502, opened 2026-06-18
Re-report UnifiedPush endpoint on tunnel connect (foss)

### LINKED ISSUE #485: google→foss in-place upgrade stuck on push_kind=fcm (no foss connect-time endpoint report) → background wake dead [pre-F-Droid blocker]
## Severity: pre-F-Droid blocker

A device upgraded **in place** from the `google` build to the `foss` build keeps its subdomain and cert and serves fine — but its relay `push_kind` stays `fcm` forever, so on-demand background wake (the core feature) is **dead**: the relay FCM-wakes a device that has no Firebase, and the wake never arrives.

## Confirmed live
On the user's Pixel 10 (`com.rousecontext`, foss, owns subdomain **`weird`**): relay record is `push_kind=fcm` and **never** went `unifiedpush` in the entire journal; relay was observed FCM-waking `weird` in response to a real AI client hitting `weird.rousecontext.com`, while the foss device (no FCM) never answered. The device otherwise authenticates and serves fine (it successfully rotated secrets on `weird`), proving identity/cert survive — only the push transport is stuck.

## What is NOT broken
Subdomain/cert/keypair identity migrate cleanly: both flavors use the same hardware Keystore key (`AndroidKeystoreDeviceKeyManager` alias `rouse_device_key`), and renewal is keyed on `subdomain` + a CSR signature against the stored public key (`relay/src/api/renew.rs handle_renew`), not on Firebase UID. So renew/rotate/WS all work post-migration. The stale `firebase_uid` on the record is harmless.

## Root cause
The foss build never reports its UnifiedPush endpoint for an **already-registered** subdomain:
- The only flip path is `app/src/foss/.../delivery/UnifiedPushBackgroundDelivery.refreshEndpoint()`, which fires only when a UnifiedPush endpoint (re)delivery coincides with a live tunnel; otherwise it defers, "relying on a future reconnect to re-report."
- That per-connect re-report **does not exist for foss**: `work/.../TunnelForegroundService.reportOnConnect()` (call sites ~205, 344) is bound to `app/src/foss/.../push/NoOpConnectPushReporter.kt` — a **no-op**. (Google binds `FcmConnectPushReporter`, which re-sends every connect.)
- So nothing ever sends `{"type":"push_endpoint","kind":"unifiedpush",...}` (the relay would accept it — `relay/src/api/ws.rs apply_push_update`, gated only by the mTLS cert) and the record stays `fcm`.

## Fix (recommended — app-side, self-healing, no subdomain change)
Give the foss flavor a real `ConnectPushReporter` (replace `NoOpConnectPushReporter`) that, on each tunnel connect, reads the persisted UnifiedPush endpoint (already stored by `UnifiedPushBackgroundDelivery` under prefs `unifiedpush_delivery`/`endpoint`) and calls `tunnelClient.sendPushEndpoint("unifiedpush", endpoint)` — mirroring `app/src/google/.../push/FcmConnectPushReporter.kt`. This flips a stuck device to `unifiedpush` on its next foreground connect (the only time a wake-dead device connects), with **no subdomain change** and **no AI-client reconfiguration**, and also covers reinstalls where the endpoint arrived while the tunnel was down. No relay change needed.

## Alternative (worse): cycle the subdomain
Re-registering recovers wake but mints a **new** subdomain — a foss re-register looks the device up by keypair thumbprint (`relay/src/api/register.rs`), and a Firebase-era record has `key_thumbprint=None`, so it's treated as brand-new (fresh subdomain, `push_kind=unifiedpush`). That forces every configured AI client to be repointed. The connect-reporter fix avoids this.

## No existing migration handling
`RegisterIdentity` keys Firebase-by-UID and keypair-by-thumbprint as disjoint namespaces; renewal works only by the coincidence that both flavors share `rouse_device_key`. The push-transport transition is an unhandled gap.

Part of #460. Found by dogfooding the foss release on a previously-google device.

monkopedia-coder:
Closes #485

## Problem
UnifiedPush endpoints can rotate at any time (a distributor may send a fresh `NEW_ENDPOINT`). The foss build only re-reported a rotated endpoint via `UnifiedPushBackgroundDelivery.refreshEndpoint` **while the tunnel was already connected**. If the endpoint rotated while the tunnel was DOWN, the new endpoint was persisted locally but `refreshEndpoint` deferred the relay update ("a future reconnect re-reports it") — and the connect-time reporter (`NoOpConnectPushReporter`) did nothing. So the relay kept waking the **stale** endpoint and the device became unwakeable until a coincidental connected redelivery. The google flavor doesn't have this gap because `FcmConnectPushReporter` re-sends the token on every connect.

## Fix
Replace the foss `NoOpConnectPushReporter` with a real `UnifiedPushConnectPushReporter` that, on each `reportOnConnect()` (fired post-connect from `TunnelForegroundService`), reads the persisted endpoint and calls `tunnelClient.sendPushEndpoint("unifiedpush", endpoint)` — mirroring the google `FcmConnectPushReporter`.

- The persisted endpoint is read via a new `UnifiedPushBackgroundDelivery.currentEndpoint()` that returns the same prefs entry the delivery writes, so there is a single source of truth for the value/key. `PUSH_KIND` is reused from the delivery companion rather than duplicated.
- DI binding in the foss `DistributionModule` now wires the real reporter.
- No relay change — the WS `push_endpoint` path already accepts it under the mTLS cert.

## Strand confirmed (confirm-then-fix)
The test reproduces the gap: endpoint rotates while the tunnel is DISCONNECTED (persisted, refresh deferred — verified nothing sent), then a connect fires `reportOnConnect()` and the relay must receive the NEW endpoint. Against the current no-op this FAILED at the `sendPushEndpoint("unifiedpush", <new>)` assertion; with the real reporter it passes.

## Verification
- `:app:compileDebugKotlin -Pgoogle` compiles (google `FcmConnectPushReporter` untouched)
- `:app:ktlintCheck` + `detekt` clean


monkopedia-reviewer (approved):
Reviewed against all five focus points — this is clean.\n\n**Idempotent:** Each connect sends the current persisted endpoint, mirroring `FcmConnectPushReporter`; the relay receiving a repeated identical endpoint is harmless, so reconnect churn is benign.\n\n**Handles rotated endpoint:** `delivery.currentEndpoint()` reads directly from SharedPrefs on each call — no in-memory cache — so a value written by `onEndpoint()` while disconnected is picked up fresh on reconnect.\n\n**No-endpoint no-op:** `isNullOrBlank()` guard covers both null (never registered) and blank; returns early with no relay call.\n\n**Google path unaffected:** `FcmConnectPushReporter` is untouched. The only hits for `NoOpConnectPushReporter` are stale agent worktrees in `.claude/worktrees/` — the main/PR tree is clean.\n\n**Test genuinely proves the strand:** Sets tunnel to `DISCONNECTED`, drives `onEndpoint(ROTATED_ENDPOINT)` + `advanceUntilIdle()`, asserts zero sends (gap confirmed), then calls `reportOnConnect()` and asserts exactly one `sendPushEndpoint("unifiedpush", ROTATED_ENDPOINT)` call (fix confirmed). The second test covers the no-endpoint path. Fail-against-old-NoOp is structural (NoOp never calls `sendPushEndpoint`).\n\nCoroutine conventions respected: `reportOnConnect()` is a short one-shot suspend call, no GlobalScope, no unscoped scope creation.\n\nCI green on all three checks (Build & Test, Lint, test). LGTM.

